PR #2, opened 2016-03-01
Working: minus variables

HookyQR:
It's Javascript internally, so would probably benefit from refactoring back to typescript.

Thanks for getting in touch.

Play, pause, stepping, and breakpoints work. 

Adding and deleting break points works ok. Can handle conditional break points only when adding. (The change test and collection doesn't note if it's conditional or not.)

runtime (rdebug-ide) args + EXE + executable args work. (Haven't added to the setup/options files)

I removed the use of xml parser, seemed a bit heavy for what is a known string structure. Replaced with individual regex's. There is some scope to DRY the code there as well, plenty of repetition.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -1,5 +1,7 @@
 # Ruby Language and Debugging Support for Visual Studio Code
 
+[![Join the chat at https://gitter.im/rebornix/vscode-ruby](https://badges.gitter.im/rebornix/vscode-ruby.svg)](https://gitter.im/rebornix/vscode-ruby?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
+
 This extension provides rich Ruby language and debugging support for Visual Studio Code.
 It's still in progress ( [GitHub](https://github.com/rebornix/vscode-ruby.git) ), please expect frequent updates with breaking changes before 1.0.
 


src/main.ts
```diff
@@ -20,32 +24,31 @@ export interface LaunchRequestArguments {
 	/** An absolute path to the program to debug. */
 	program: string;
 	/** Automatically stop target after launch. If not specified, target does not stop. */
-	stopOnEntry?: boolean;
+	stopOnEntry ? : boolean;
```
rebornix: Let's keep it with what people does in VS Code code base, use `stopOnEntry?` with no inner spaces

src/main.ts
```diff
@@ -20,32 +24,31 @@ export interface LaunchRequestArguments {
...
+class RubyDebugSession extends DebugSession {
 
 	// we don't support multiple threads, so we can use a hardcoded ID for the default thread
 	private static THREAD_ID = 1;
 
-	private _breakpointId = 1000;
-
 	// maps from sourceFile to array of Breakpoints
-	private _breakPoints = new Map<string, DebugProtocol.Breakpoint[]>();
+	private _breakPoints = new Map < string, DebugProtocol.Breakpoint[] > ();
```
rebornix: No surrounded spaces.

src/main.ts
```diff
@@ -62,209 +65,233 @@ class MockDebugSession extends DebugSession {
 	 * The 'initialize' request is the first request called by the frontend
 	 * to interrogate the features the debug adapter provides.
 	 */
-	protected initializeRequest(response: DebugProtocol.InitializeResponse, args: DebugProtocol.InitializeRequestArguments): void {
-		// This debug adapter implements the configurationDoneRequest.
-		response.body.supportsConfigurationDoneRequest = true;
+	protected initializeRequest(response: DebugProtocol.InitializeResponse, args: DebugProtocol.InitializeRequestArguments):
+		void {
```
rebornix: Keep it at the end of previous line?

rebornix:
@HookyQR Thanks for your input, your pr looks great !

Just one more comment, for coding style, let's follow VS Code [Coding Guidelines](https://github.com/Microsoft/vscode/wiki/Coding-Guidelines) and what people do in VS Code code base, then we can have a consistent style between us.

rebornix:
@HookyQR I've made some new changes to the code base, fix some bugs, add new features. It might have some common parts with your original PR (like deleting breakpoints, etc), but I really look forward to your update :+1:  

BTW, I've added some [installation instruction](https://github.com/rebornix/vscode-ruby#install-ruby-dependencies), feel free to drop your lines if there is anything missing.

HookyQR:
@rebornix Since my code as diverged from the current master, I'll do another fork this weekend and start from there. There are a couple of things I've seen already, hopefully I can get a new PR up before your code base changes significantly.

rebornix:
@HookyQR there won't be any new breaking change to the source code this weekend, so go ahead!

I'll spend some time on writing tests and fix corner cases to make sure I won't break anything when adding new features.

PR #5, opened 2016-03-06
Working debugger.

HookyQR:

- [x] Build pass!

There will be some js -> ts changes required. 

## Current state:
* [x] Runtime (debugger) arguments accepted
* [x] Program arguments accepted (as `"args"`, just like node debugger)
* [x] Threads
* [x] Variables: 
  - [x] local
  - [x] global
  - [x] instance variable tree
* [x] Stack trace
* [x] Breakpoints
* [x] Catchpoints (break on exception) - All exceptions only
* [x] Correctly 'pause on start'
* [x] Correctly run when not 'pause on start' after breakpoints added
* [ ] Conditional breaks (Existing issues with rdebug-ide)
* [ ] Step out (Uses `finish` which has odd and inconsistent results)
* [ ] Evaluate:
  - [ ] watch: runs, but has odd results, should tie to variablesReference store
  - [ ] console line: as above
  - [ ] Optionally: send console line to debug target when STDIN is waiting

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -63,11 +63,11 @@ class RubyDebugSession extends DebugSession {
 		this.rubyProcess = new RubyProcess(args);
 
 		this.rubyProcess.on('debuggerConnect', () => {
-			this.sendEvent(new InitializedEvent());
 			this.sendResponse(response);
+			this.sendEvent(new InitializedEvent());
 		}).on('debuggerComplete', () => {
 			this.sendEvent(new TerminatedEvent());
-		}).on('exeutableOutput', (data: Buffer) => {
+		}).on('executableOutput', (data: Buffer) => {
 			this.sendEvent(new OutputEvent(data.toString(), 'stdout'));
 		}).on('executableStdErr', (error: Buffer) => {
 			this.sendEvent(new OutputEvent(error.toString(), 'stderr'));
@@ -82,7 +82,8 @@ class RubyDebugSession extends DebugSession {
 				this.sendEvent(new StoppedEvent('step', result.threadId));
 			this._hasStopped = true;
 		}).on('exception', result => {
-			this.sendEvent(new StoppedEvent('exception', result.threadId, result.messsage));
+			this.sendEvent(new OutputEvent("\nException raised: ["+result.type+"]: "+result.message+"\n",'stderr'));
+			this.sendEvent(new StoppedEvent('exception', result.threadId, result.type+": "+result.message));
 		}).on('terminalError', (error: string) => {
 			this.sendEvent(new OutputEvent("Debugger terminal error: "+ error))
 			this.sendEvent(new TerminatedEvent());
@@ -98,8 +99,24 @@ class RubyDebugSession extends DebugSession {
 	// Executed after all breakpints have been set by VS Code
 	protected configurationDoneRequest(response: DebugProtocol.ConfigurationDoneRequest, args:
 	DebugProtocol.ConfigurationDoneArguments): void {
+		this.rubyProcess.Run('start');
 	}
+	protected setExceptionBreakPointsRequest(response: DebugProtocol.SetExceptionBreakpointsResponse, args: DebugProtocol.SetExceptionBreakpointsArguments): void {
+		if (args.filters.indexOf('all') >=0){
+			//Exception is the root of all (Ruby) exceptions - this is the best we can do
+			//If someone makes their own exception class and doesn't inherit from
+			//Exception, then they really didn't expect things to work properly
+			//anyway.
 
+			//We don't do anything with the return from this, but we
+			//have to add an expectation for the output.
+			this.rubyProcess.Enqueue('catch Exception').then(()=>1);
+		}
+		else {
+			this.rubyProcess.Run('catch off');
+		}
+		this.sendResponse(response);
+	}
 	protected setBreakPointsRequest(response: DebugProtocol.SetBreakpointsResponse, args: DebugProtocol.SetBreakpointsArguments): void {
 		var path = args.source.path;
 
--- a/src/main.ts
+++ b/src/main.ts
@@ -71,7 +71,7 @@ class RubyDebugSession extends DebugSession {
 		}).on('executableStdErr', (error: Buffer) => {
 			this.sendEvent(new OutputEvent(error.toString(), 'stderr'));
 		}).on('nonTerminalError', (error: string) => {
-			this.sendEvent(new OutputEvent("Debugger error: "+error+ '\n', 'stderr'));
+			this.sendEvent(new OutputEvent("Debugger error: " + error + '\n', 'stderr'));
 		}).on('breakpoint', result => {
 			this.sendEvent(new StoppedEvent('breakpoint', result.threadId));
 		}).on('suspended', result => {
@@ -81,10 +81,10 @@ class RubyDebugSession extends DebugSession {
 				this.sendEvent(new StoppedEvent('step', result.threadId));
 			this._firstSuspendReceived = true;
 		}).on('exception', result => {
-			this.sendEvent(new OutputEvent("\nException raised: ["+result.type+"]: "+result.message+"\n",'stderr'));
-			this.sendEvent(new StoppedEvent('exception', result.threadId, result.type+": "+result.message));
+			this.sendEvent(new OutputEvent("\nException raised: [" + result.type + "]: " + result.message + "\n",'stderr'));
+			this.sendEvent(new StoppedEvent('exception', result.threadId, result.type + ": " + result.message));
 		}).on('terminalError', (error: string) => {
-			this.sendEvent(new OutputEvent("Debugger terminal error: "+ error))
+			this.sendEvent(new OutputEvent("Debugger terminal error: " + error))
 			this.sendEvent(new TerminatedEvent());
 		});
 
@@ -129,9 +129,9 @@ class RubyDebugSession extends DebugSession {
 			Promise.all(deletePromises).then(() => this._breakPoints.delete(path));
 		}
 		var breakpointAddPromises = args.breakpoints.map(b=>{
-			//let conditionString =  b.condition ? ' if '+b.condition : '';
-			//return this.rubyProcess.Enqueue('break '+ path+':'+b.line+conditionString);
-			return this.rubyProcess.Enqueue('break '+ path+':'+b.line);
+			//let conditionString =  b.condition ? ' if ' + b.condition : '';
+			//return this.rubyProcess.Enqueue('break ' + path + ':' + b.line + conditionString);
+			return this.rubyProcess.Enqueue('break ' + path + ':' + b.line);
 		});
 		Promise.all(breakpointAddPromises).then( values => {
 			let breakpoints = values.map(attributes => {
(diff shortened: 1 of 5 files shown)


HookyQR:
**Note:** this will likely break many tests as the point of running the xml parser has changed. I still think this should be replaced, not because it's bad, but because the processor has prior knowledge of the structure of the response, AND the response for `eval` is breaking the system. (Hack fix in ruby.js lines 148-171, plus the error catchers at the head of the file)

rebornix:
I really love your idea of changing XMLDom to regex as it breaks often when parsing string to XML. Let me take some time reading your code.

HookyQR:
Turns out regex won't do as I hit an unspecified limit when running against the global variables. Could only catch the inner elements with an input string of <= 1500 bytes.

XMLDom is working for now, with the crutch I dropped in for the poorly formatted eval results.

src/main.ts
```diff
@@ -18,11 +18,12 @@ import {SocketClientState} from './common';
 
 class RubyDebugSession extends DebugSession {
 
-	// we don't support multiple threads, so we can use a hardcoded ID for the default thread
-	private static THREAD_ID = 2;
-
 	private _breakpointId = 1000;
+	private _threadId = 2;
+	private _frameId = 0;
+	private _hasStopped = false;
```
rebornix: This variable can be retrieved from RubyProcess.state ?

src/main.ts
```diff
@@ -51,128 +52,105 @@ class RubyDebugSession extends DebugSession {
...
-		}
 	}
 
 	// Executed after all breakpints have been set by VS Code
-	protected configurationDoneRequest(response: DebugProtocol.ConfigurationDoneResponse, args:
+	protected configurationDoneRequest(response: DebugProtocol.ConfigurationDoneRequest, args:
 	DebugProtocol.ConfigurationDoneArguments): void {
-		var command = ['start'];
-		this.rubyProcess.Run(command.join(' ') + '\n');
+		this.rubyProcess.Run('start');
```
rebornix: It's `DebugProtocol. ConfigurationDoneResponse ` actually, and we need to send back the response.

src/main.ts
```diff
@@ -51,128 +52,105 @@ class RubyDebugSession extends DebugSession {
...
-		var linesToRemove: number[] = [];
-		var linesToUpdate: number[] = linesToAdd;
-
-		if (registeredBks) {
-			linesToRemove = registeredBks.map(b => b.line).filter(oldLine => linesToAdd.indexOf(oldLine) === -1);
-		    linesToUpdate = registeredBks.map(b => b.line).filter(oldLine => linesToAdd.indexOf(oldLine) >= 0);
+		//to ensure that breakpoints with altered conditions are set, it is
+		//best to remove all and add them back.
+		if (this._breakPoints.get(path)) {
+			this.rubyProcess.Run('delete');
```
rebornix: `delete` command will delete all breakpoints of all files, in this case, if we are setting breakpoints for two files, a second `delete` will delete the breakpoints we set for previous file, so keep `linesToRemovePromise`

src/main.ts
HookyQR: `_hasStopped` is if `.Run('start')` has been called. It enables display of `stopped on entry` instead of a standard step, which doesn't show anything. So it's not the state of the socket (which `.state` tracks), rather whether or not a suspended has already been received.

src/main.ts
```diff
@@ -51,128 +52,105 @@ class RubyDebugSession extends DebugSession {
...
-
-		var linesToRemovePromise = linesToRemove.map(line => {
-			var bk = registeredBks.filter(b=> b.line === line)[0];
-			return that.rubyProcess.Enqueue('delete ' + bk.id + '\n');
+		var linesToUpdatePromise = args.breakpoints.map(b=>{
+			let conditionString =  b.condition ? ' if '+b.condition : '';
+			return this.rubyProcess.Enqueue('break '+ path+':'+b.line+conditionString);
 		});
+		Promise.all(linesToUpdatePromise).then( values => {
+			let breakpoints = values.map(attributes => {
```
rebornix: seems that the condition command doesn't work? If so let's don't add it here.

src/main.ts
```diff
@@ -51,128 +52,105 @@ class RubyDebugSession extends DebugSession {
...
-			}
-
-			var threads = new Array<Thread>();
-			for(let i= 0; i < xml.documentElement.childNodes.length; i++) {
-				var threadNode = xml.documentElement.childNodes.item(i);
-				var threadId = threadNode.attributes.getNamedItem('id');
-
-				threads.push(new Thread(+threadId.value, 'thread_'+threadId.value));
-			}
+		this.rubyProcess.Enqueue('thread list\n').then(results => {
```
rebornix: `\n` is redundant now in your PR.

src/main.ts
```diff
@@ -183,82 +161,79 @@ class RubyDebugSession extends DebugSession {
 		The request returns a stacktrace from the current execution state.
 	*/
 	protected stackTraceRequest(response: DebugProtocol.StackTraceResponse, args: DebugProtocol.StackTraceArguments): void {
-		var that = this;
-
-		this.rubyProcess.Enqueue('where\n').then((xml: XMLDocument) => {
-			if (xml.documentElement.nodeName !== 'frames') {
-				return;
-			}
+		this.rubyProcess.Enqueue('where\n').then(results => {
```
rebornix: same here, we already have `\n` in `Enqueue`?

src/main.ts
```diff
@@ -266,118 +241,65 @@ class RubyDebugSession extends DebugSession {
...
 	}
 
-	/** Variables request; value of command field is 'variables'.
-		Retrieves all children for the given variable reference.
-	*/
 	protected variablesRequest(response: DebugProtocol.VariablesResponse, args: DebugProtocol.VariablesArguments): void {
 		var varRef = this._variableHandles.get(args.variablesReference);
+		let varPromise;
+		if ( varRef.objectId ){
+			varPromise = this.rubyProcess.Enqueue('var i '+varRef.objectId).then(results => this.createVariableReference(results));
```
rebornix: we may want to have space around `+`

src/ruby.ts
```diff
@@ -32,200 +46,199 @@ export class RubyProcess extends EventEmitter {
...
-			runtimeExecutable = 'rdebug-ide';
-		}
 		else {
 			// platform: linux
 			runtimeExecutable = 'rdebug-ide';
 		}
 
-		var programArgs = [];
 		var processCwd = dirname(this.launchArgs.program);
+		//xd only if showDebuggerOutput is set - it is also a runtime arg, not a program arg
```
rebornix: for now, our runtime is `xdebug-ide` while the runtime should be `ruby` actually, but you can leave it there and let me fix it later.

src/main.ts
rebornix: it makes sense. maybe we can give it another name with less confusion.

rebornix:
As you already handled poorly formatted rdebug-ide result then XMLDom is good enough for current phase.

The PR looks good to me, just give some tiny changes against my comments and make sure `npm test` pass ( it only contains debugger launch, breakpoint hit and variable inspection, so if the cases don't pass, there must be some significant bug in the PR), then I'm gonna to merge it!

HookyQR:
Will fix up all these changes now.

src/main.ts
HookyQR: is now `_firstSuspendReceived`

src/main.ts
HookyQR: Oops. Reverted, and responded

src/main.ts
HookyQR: Now deletes only those in the map for that file

src/main.ts
HookyQR: Still adding the breakpoint here, even if a condition is set. Best to stipulate the conditions aren't honoured in docs.
Have left the code for it in, but commented out, for when it does work.
Also, renamed the promises to `breakpointAddPromises`

src/main.ts
HookyQR: Done

src/main.ts
HookyQR: done

src/ruby.ts
HookyQR: I consider the runtime to be the debugger, since we aren't passing arguments directly to ruby. Note that the `d` flag (line tracing) doesn't work anyway.

Added some explanation around this.

HookyQR:
Ran tests. Worked fine my end.

rebornix:
It was the travis config issue, just ignore it, i'll fix it tonight

src/ruby.ts
rebornix: FYI, the command ran in RubyMine is `/usr/bin/ruby -e at_exit{sleep(1)};$stdout.sync=true;$stderr.sync=true;load($0=ARGV.shift) /Users/rebornix/.gem/ruby/2.0.0/gems/ruby-debug-ide-0.4.32/bin/rdebug-ide --disable-int-handler --evaluation-timeout 10 --rubymine-protocol-extensions --port 55354 --dispatcher-port 55355 -- /Users/rebornix/Documents/code/testrb/test.rb` , I suppose we might use Ruby runtime someday, but for now let's stick with rdebug-ide.

rebornix:
I tried to merge your pr but travis build failed and on my devbox, when I try to `tsc -p ./src`, it outputs

```
src/ruby.ts(74,52): error TS2345: Argument of type 'string[]' is not assignable to parameter of type 'string'.
src/ruby.ts(153,22): error TS2339: Property 'includes' does not exist on type 'string'.
src/ruby.ts(179,22): error TS2339: Property 'threadId' does not exist on type '{}'.
src/ruby.ts(179,44): error TS2339: Property 'threadId' does not exist on type '{}'.
src/ruby.ts(179,67): error TS2339: Property 'threadId' does not exist on type '{}'.
src/ruby.ts(197,62): error TS2339: Property 'tagName' does not exist on type 'Node'.
src/ruby.ts(198,21): error TS2339: Property 'tagName' does not exist on type 'Node'.
src/main.ts(222,88): error TS2345: Argument of type '{ objectId: any; kind: any; }' is not assignable to parameter of type 'IDebugVariable'.
  Object literal may only specify known properties, and 'objectId' does not exist in type 'IDebugVariable'.
src/main.ts(239,34): error TS2346: Supplied parameters do not match any signature of call target.
src/main.ts(240,35): error TS2346: Supplied parameters do not match any signature of call target.
src/main.ts(251,15): error TS2339: Property 'objectId' does not exist on type 'IDebugVariable'.
src/main.ts(252,60): error TS2339: Property 'objectId' does not exist on type 'IDebugVariable'.
```

it seems ES6 issues, but even if I change the TypeScript compilerOptions `target` to `ES6`, the error still exists. So I just revert this change to unlock further development. 

Would you mind share your local development setting (maybe compilerOptions or your TypeScript version) to make it pass and re-submit a PR? Thanks!

HookyQR:
None of those errors cause failure.

I've reduced it to 4 'errors' by changing some of the signatures and defining types as `:any` in some cases. The remaining erros are:

```
src/ruby.ts(74,52): error TS2345: Argument of type 'string[]' is not assignable to parameter of type 'string'.
src/ruby.ts(153,22): error TS2339: Property 'includes' does not exist on type 'string'.
src/main.ts(239,34): error TS2346: Supplied parameters do not match any signature of call target.
src/main.ts(240,35): error TS2346: Supplied parameters do not match any signature of call target.
```

* The first one is because typescript doesn't think you can `concat` an array.
* The second one says `include` isn't a function on `string` which it is. You can change this to `.indexOf(...) >= 0` to get rid of this error.
* The last two are oddly saying that a number isn't a number? It is complaining about the signature of `_variableHandles.create()` doesn't match the type of the second argument of `new Scope()`. I don't know why it's doing that.

Passes tests fine. I haven't changed your compile settings.

Have added another pull request.

PR #6, opened 2016-03-06
Same PR, less 'errors'

HookyQR:

- [x] Build pass!

As per last PR.

None of those errors cause failure.

I've reduced it to 4 'errors' by changing some of the signatures and defining types as `:any` in some cases. The remaining erros are:

```
src/ruby.ts(74,52): error TS2345: Argument of type 'string[]' is not assignable to parameter of type 'string'.
src/ruby.ts(153,22): error TS2339: Property 'includes' does not exist on type 'string'.
src/main.ts(239,34): error TS2346: Supplied parameters do not match any signature of call target.
src/main.ts(240,35): error TS2346: Supplied parameters do not match any signature of call target.
```

* The first one is because typescript doesn't think you can `concat` an array.
* The second one says `include` isn't a function on `string` which it is. You can change this to `.indexOf(...) >= 0` to get rid of this error.
* The last two are oddly saying that a number isn't a number? It is complaining about the signature of `_variableHandles.create()` doesn't match the type of the second argument of `new Scope()`. I don't know why it's doing that.

Passes tests fine. I haven't changed your compile settings.

Have added another pull request.

### DIFF
--- a/src/ruby.ts
+++ b/src/ruby.ts
@@ -12,14 +12,28 @@ import {DOMParser} from 'xmldom';
 import {LaunchRequestArguments, IRubyEvaluationResult, IDebugVariable, ICommand} from './interface';
 import {SocketClientState} from './common';
 
+var domErrorLocator: any = {};
+
 export class RubyProcess extends EventEmitter {
 	private debugSocketClient : net.Socket = null;
 	private buffer: string;
+	private socketConnected: boolean;
 	private parser: DOMParser;
 	private debugprocess: childProcess.ChildProcess;
 	private launchArgs: LaunchRequestArguments;
-	private pendingCommands: ICommand[];
+	private pendingResponses: ICommand[];
+	private pendingCommands: any[];
 	private _state: SocketClientState;
+
+	private domErrors: any;
+	private domErrorHandler(type: string, error: string){
+		this.domErrors.push({
+			lineNumber: domErrorLocator.lineNumber,
+			columnNumber: domErrorLocator.columnNumber,
+			error: error,
+			type: type
+		});
+	}
 
 	get state(): SocketClientState {
 		return this._state;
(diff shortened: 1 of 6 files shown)


HookyQR:
Travis seems to be failing on the compile because of these `errors`.

It feels wrong to be changing valid JS just to please a compiler's pedantry.

HookyQR:
First two errors removed.

HookyQR:
All errors removed.

HookyQR:
Again, tests passed here. I haven't changed anything in the setups.

rebornix:
I pulled your repo to my devbox ( Mac ) and run `npm test` but the last two test cases failed, I'll take a look at your code or add some detailed failing information in the test cases

src/ruby.ts
```diff
@@ -32,200 +46,199 @@ export class RubyProcess extends EventEmitter {
...
 
 		this.debugSocketClient = new net.Socket( {
 			type: 'tcp4'
 		});
-
 		this.debugSocketClient.on('connect', (buffer: Buffer) => {
-			that.state = SocketClientState.connected;
-			that.emit('debuggerConnect');
+			this.state = SocketClientState.connected;
+			//first thing we have to send is the start - if stopOnEntry is
```
rebornix: Here is the catch. If you don't set any breakpoint for the source code, once you triggered  `start` command, VS Code Debug UI or our test code will never receive `initialized` event.

rebornix:
The workflow of VS Code debugger initialization and invocation is likely to be:

1. VS Code call DebugSession constructor
2. VS Code call DebugSession initializeRequest, we send back response
3. VS Code call DebugSession LaunchRequest or AttachRequest, we create a process for `rdebug-ide`, once we successfully connect to the socket, we send back response and send out an `InitializedEvent`.
4. VS Code call `setBreakpointsRequest`, we send `break xxx:line` to `rdebug-ide`
5. VS Code call `configurationDoneRequest`, then we run `start` command to `rdebug-ide` and start debugging our code.

Step 4 and 5 can only be triggered after the socket is successfully connected to the debugger, not before.

src/ruby.ts
HookyQR: So it turns out the problem was a copy and paste error when I was fixing the 'errors' that were reported in the type scrypt build. The problem is nothing to do with what you've said here.

The problem was in the main.js file in the scopes request where I had put `results[1]` for both.

I'm fixing the merge now an will push back up when it's running. Not sure it will work with any new test.

HookyQR:
Travis now passing. Turns out it was the node version the tester was using. The version in VS Code is 4.1.1, you were using 0.12.7 which I guess doesn't have some functionality we are requiring.

rebornix:
Yup, Travis seems Okay now, but there still some issues with Appveyor, give me sometime to take a look.

HookyQR:
I've already pushed a change for Appveyor. Waiting for it to complete.

HookyQR:
BAM!!! First test passed!

rebornix:
Everything is okay now! Perfect!

Any idea about why there are 42 commits in your PR? Maybe next time you can do some rebasing before sending out new change, but maybe this time you can squash your change into several few commits.

HookyQR:
I've got my upstream source as this project and the target as my project, so a couple of times the merge has back tracked and re-applied the whole project. Would obviously be easier if I could run a branch here, but I'll let you make a call on that.

rebornix:
@HookyQR you are collabrator of this repo now, so you can push your local rebased branch here. PR is still expected then we can know whether it break anything.

HookyQR:
Thanks @rebornix. What do you want done with this PR? Are you going to pull it in now? Otherwise can you send me a specific commit you want it rebased from. It's the end of the night for me, so it will have to be tomorrow night though.

rebornix:
You can squash your change to a few commits, do a rebase to commit 90756f67998965296647d90073a9c868d2d72db6. Then push forcely to your fork, then it will take effect in this pr.

Take your time and have a good night, I'll merge it once your pr is ready and release 0.2.0.

HookyQR:
Single commit to be merged now. @rebornix all yours.

PR #23, opened 2016-05-16
Merged the language support from groksrc/ruby_colors_for_vscode.

lextm:
It would be ideal to have all Ruby support (language and debugger) in the same extension, so I manually copied the bits from Drew Cain's repo. He hasn't updated the repo since February, so maybe later the language support can be maintained by vscode-ruby owners for Ruby users in VS Code.

The original code was released under MIT, which is the same as this repo.


rebornix:
@lextm thanks for your input :+1:

groksrc:
@lextm has my thanks as well. My effort was really just to bootstrap ruby for the editor. I use `vscode-ruby` as my day-to-day extension for ruby now. 👍

PR #37, opened 2016-06-14
add decreaseIndentPattern

rebornix:
- [x] Build pass!

Add `decreaseIndentPattern` to have keywords like `end` get correct auto unindent.

This is the code change in our side while seems VS Code doesn't take `decreaseIndentPattern` correctly as Atom or Sublime does, we'll wait to see when VS Code gets a fix for it.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -9,7 +9,7 @@
 			"args": [
 				"--extensionDevelopmentPath=${workspaceRoot}"
 			],
-			"sourceMaps": true,
+			"sourceMaps": false,
 			"outDir": "${workspaceRoot}/out"
 		},
 		{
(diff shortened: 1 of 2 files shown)


chriskacerguis:
Anyway that we can get this Merged?  Looks like it will solve my issue https://github.com/rubyide/vscode-ruby/issues/45

rebornix:
@chriskacerguis like I mentioned above, it depends on Code's decreaseIndent support so I suppose merging this PR won't work directly.

chriskacerguis:
@rebornix I have https://github.com/Microsoft/vscode/issues/10507 open with VScode as well, and they are saying it's this extension not supporting decreaseIndentPattern.  So, this won't "fix" the issue, but at least it will be in there.

Also, I got your email, I'll reply there :)

rebornix:
@chriskacerguis Look forward to your email!

PR #41, opened 2016-07-28
Add completion feature

mattn:

- [x] Build pass!

I wonder that you will have an interesting.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -107,6 +107,18 @@ Settings available (in your VSCode workspace) for each of the linters:
 }
 ```
 
+## Completion
+
+You will need to install the ruby gem for Intellisense/Completion.
+
+* rcodetools
+
+You need to restart vscode after installation of rcodetools. Then type CTRL-Space after the leading words like below.
+
+```ruby
+[1,2,3].e #<= Type CTRL-Space at here
+```
+
 ## Features
 
 - Ruby scripts debugging
--- a/readme.md
+++ b/readme.md
@@ -130,6 +130,7 @@ To enable method completion in ruby: `gem install rcodetools`. You may need to r
   * Breaking on uncaught exceptions and errors
   * Attach requests
   * Breakpoints can also be set in `.erb` files
+  * IntelliSense and autocomplete
 
 - Unit/Integration tests debugging
   * RSpec
@@ -150,7 +151,6 @@ To enable method completion in ruby: `gem install rcodetools`. You may need to r
 - Rake
 - Gem
 - IRB console
-- IntelliSense and autocomplete
 
 ## Main contributors
 
--- a/readme.md
+++ b/readme.md
@@ -107,16 +107,12 @@ Settings available (in your VSCode workspace) for each of the linters:
 }
 ```
 
-## Completion
+## Autocomplete
 
-You will need to install the ruby gem for Intellisense/Completion.
-
-* rcodetools
-
-You need to restart vscode after installation of rcodetools. Then type CTRL-Space after the leading words like below.
+To enable method completion in ruby: `gem install rcodetools`. You may need to restart Visual Studio Code the first time.
 
 ```ruby
-[1,2,3].e #<= Type CTRL-Space at here
+[1, 2, 3].e #<= Press CTRL-Space here
 ```
 
 ## Features
--- a/ruby.js
+++ b/ruby.js
@@ -111,8 +111,6 @@ function completionProvider(document, position, token) {
 	return new Promise((resolve, reject) => {
 		const line = position.line + 1;
 		const column = position.character;
-		console.log("line=" , line);
-		console.log("column=" , column);
 
 		var sh, flag;
 		if (process.platform === 'win32') {
--- a/ruby.js
+++ b/ruby.js
@@ -1,6 +1,9 @@
 "use strict";
 let vscode = require('vscode');
 let linters = require('./lint/lint');
+let cp = require('child_process');
+let which = require('which');
+
 const severities = {
 	refactor: vscode.DiagnosticSeverity.Hint,
 	convention: vscode.DiagnosticSeverity.Information,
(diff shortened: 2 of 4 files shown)


ruby.js
```diff
@@ -104,6 +107,54 @@ function balanceEvent(event) {
 	if (event && event.document) balancePairs(event.document);
 }
 
+function completionProvider(document, position, token) {
+	return new Promise((resolve, reject) => {
+		const line = position.line + 1;
+		const column = position.character;
+		console.log("line=" , line);
```
rebornix: you may want to remove this.

rebornix:
@mattn thanks for your contribution, it look good to me!

Do you mind adding a simple guidance in README after `Lineters` or putting the setup info in Wiki and adding a link in README? I don't use `rct-complete` so it will be awesome if there is a doc for it and it will be usefull to users as well.

mattn:
Added small installation instruction for the completion feature. But I'm not native speaker, if you feel wrong, please let me known.

rebornix:
@mattn thanks again! The introduction is neat! I'm not a native speaker as well, let's release it and see people's feedback!

HookyQR:
@mattn I'd like to see this with the shell dependancy. You don't need that in the call. It can affect the which (or any) version of the processor is called if you're using rbenv (or rvm). Not sure how `which` will work with them either.

Also, adding a filter on the retuned strings to get rid of the methods that aren't really completions. (Operators and such) with:
```
if (/^[^\w]/.test(items[0])) return;
```

Adding a completionItemKind as the second element to the CompletionItem constructor would be good. Method is probably best.

Change the `filterText` to be on `items[0]` rather than [1], otherwise it uses the source class/module prefix of the second part when picking a 'best' result.

There are a couple more `console` calls in there that should be removed.

mattn:
@HookyQR Thanks for your review. Fixed.

orta:
Wow, awesome, I had booked today to look at building this same thing - thanks @mattn 👍

readme.md
```diff
@@ -107,6 +107,18 @@ Settings available (in your VSCode workspace) for each of the linters:
 }
 ```
 
+## Completion
+
+You will need to install the ruby gem for Intellisense/Completion.
+
+* rcodetools
+
+You need to restart vscode after installation of rcodetools. Then type CTRL-Space after the leading words like below.
```
orta: To make this read better in English, maybe try this?

```
## Intellisense/Completion

To have completion in ruby, you will need to install the gem `rcodetools` via `gem install rcodetools`. For the first time, you will need to restart VS Code. For an example of it in action, try writing:

```ruby
[1, 2, 3].e #<= Press CTRL-Space here
```

ruby.js
```diff
@@ -104,6 +107,46 @@ function balanceEvent(event) {
...
+		const column = position.character;
+
+		var child = cp.spawn(completeCmd, [
+			'--completion-class-info',
+			'--dev',
+			'--fork',
+			'--line='+line,
+			'--column='+column], { detached: false });
+		child.stderr.on('data', (data) => {
+			console.log("stderr:", data.toString());
```
HookyQR: Drop this. If it's not reported to the user interface, there's no need to report it.

ruby.js
```diff
@@ -104,6 +107,46 @@ function balanceEvent(event) {
...
+			reject(data);
```
HookyQR: Move this to `child.stderr.on('end', ...` instead. Avoids double calling reject. (It's unlikely to happen, but possible)

ruby.js
```diff
@@ -104,6 +107,46 @@ function balanceEvent(event) {
...
+		});
+		child.stdout.on('data', (data) => {
+			var completionItems = [];
```
HookyQR: Collect the completionItems in the scope, but handle the response in `.on('end'` instead.

ruby.js
```diff
@@ -104,6 +107,46 @@ function balanceEvent(event) {
...
+				completionItem.label = items[0];
+				completionItem.kind = vscode.CompletionItemKind.Method;
+				completionItems.push(completionItem);
+			}, this);
+			if (completionItems.length == 0)
+				return reject([]);
+			return resolve(completionItems);
+		});
+		child.stdin.write(document.getText());
+		child.stdin.end();
```
HookyQR: Collapse these calls into `child.stdin.end(document.getText())`.

ruby.js
```diff
@@ -104,6 +107,46 @@ function balanceEvent(event) {
...
+	return new Promise((resolve, reject) => {
+		const line = position.line + 1;
+		const column = position.character;
+
+		var child = cp.spawn(completeCmd, [
+			'--completion-class-info',
+			'--dev',
+			'--fork',
+			'--line='+line,
+			'--column='+column], { detached: false });
```
HookyQR: `detached: false` is the default.

ruby.js
```diff
@@ -157,6 +200,14 @@ function activate(context) {
 		balancePairs(vscode.window.activeTextEditor.document);
 	}
 
+	which("rct-complete", function(err, found) {
```
HookyQR: I'm not really a fan of this call because I use `rbenv`. I'd rather it just assume it is in the path, or have a setting in the users config to point to it.

Also, this may cause a problem if the user installs it for one version of ruby, but the shell that is spawned for `which` can't find it.

I guess it's ok like this until the issues start coming in. If they don't, then all is well. 😉

readme.md
```diff
@@ -107,6 +107,18 @@ Settings available (in your VSCode workspace) for each of the linters:
...
+
+```ruby
+[1,2,3].e #<= Type CTRL-Space at here
```
HookyQR: `Press CTRL-Space here`

ruby.js
mattn: Thanks for your review. I don't like handling completion event even though rct-cmplete is not installed. How about to try to run `rct-complete --help` at once to check rct-complete is installed?

mattn: Ah, I just remember why I did. `cp.spawn` doesn't accept only command name on windows.

mattn: Sorry, I figure out this problem. I'm using msys2 mingw64 environment. And ruby on mingw64 doesn't allow non-absolute path in the command.

mattn:
Fixed problems @HookyQR suggested. Thanks.

ruby.js
```diff
@@ -104,6 +106,51 @@ function balanceEvent(event) {
...
+
+		var child = completeCommand([
+			'--completion-class-info',
+			'--dev',
+			'--fork',
+			'--line='+line,
+			'--column='+column]);
+
+		var outbuf = [], errbuf = [];
+		child.stderr.on('data', (data) => errbuf.push(data.toString()));
```
HookyQR: Push the buffer rather than the string.

ruby.js
```diff
@@ -104,6 +106,51 @@ function balanceEvent(event) {
...
+		child.stdout.on('data', (data) => outbuf.push(data.toString()));
+		child.stdout.on('end', () => {
+			if (errbuf.length > 0) reject(errbuf.join(''));
+
+			var completionItems = [];
+			outbuf.join('').split('\n').forEach(function(elem) {
```
HookyQR: Then call `Buffer.concat(outbuf).toString().split ...` here.

ruby.js
```diff
@@ -104,6 +106,51 @@ function balanceEvent(event) {
...
+			'--dev',
+			'--fork',
+			'--line='+line,
+			'--column='+column]);
+
+		var outbuf = [], errbuf = [];
+		child.stderr.on('data', (data) => errbuf.push(data.toString()));
+		child.stdout.on('data', (data) => outbuf.push(data.toString()));
+		child.stdout.on('end', () => {
+			if (errbuf.length > 0) reject(errbuf.join(''));
```
HookyQR: This will still cause a double call of one of the promise returns. If you really want to handle the `stderr` output this way, you'll have to `return` from here.

ruby.js
```diff
@@ -104,6 +106,51 @@ function balanceEvent(event) {
...
+	return new Promise((resolve, reject) => {
+		const line = position.line + 1;
+		const column = position.character;
+
+		var child = completeCommand([
+			'--completion-class-info',
+			'--dev',
+			'--fork',
+			'--line='+line,
+			'--column='+column]);
```
HookyQR: I like this setup. Nice work around.

mattn:
used Buffer.

mattn:
Ah, sorry. I just notice what you mean. Will fix.

rebornix:
@mattn sorry for the late merging. The feature is insanely amazing, thanks for it!

HookyQR:
Hey @rebornix, can you package and release a new version with this please. I'm still seeing `0.5.2` as the active release.

rebornix:
@HookyQR It's shiped as 0.5.3 with another PR merged.

PR #47, opened 2016-08-25
Add autoClosingPairs and surroundingPairs

karuna:
Add autoClosingPairs and surroundingPairs

- [x] Build pass!


ruby.js
```diff
@@ -42,8 +42,25 @@ const langConfig = {
...
+		["{", "}"],
+		["[", "]"],
+		["(", ")"],
+		["\"", "\""],
+		["'", "'"],
+		["`", "`"]
+	],
 	indentationRules: {
-		increaseIndentPattern: /^\s*((begin|class|def|else|elsif|ensure|for|if|module|rescue|unless|until|when|while)|(.*\sdo\b))\b[^\{;]*$/
+		increaseIndentPattern: /^\s*((after|before|begin|class|context|def|describe|else|elsif|ensure|for|if|it|module|rescue|unless|until|when|while)|(.*\sdo\b))\b[^\{;]*$/,
```
HookyQR: `after`, `before` and `describe` are rspec methods, not reserved words. The indentation requirements for them should be caught with the `do` or `{` indenters.

package.json
```diff
@@ -178,7 +178,8 @@
 			"id": "ruby",
 			"aliases": ["Ruby", "ruby"],
 			"extensions": [".rb", ".rbx", ".rjs", ".Rakefile", ".rake", ".cgi", ".fcgi", ".gemspec", ".irbrc", ".capfile",
-				".ru", ".prawn", ".Gemfile", ".Guardfile", ".Vagrantfile"]
+				".ru", ".prawn", ".Gemfile", ".Guardfile", ".Vagrantfile"],
+			"configuration": "./language-configuration.json"
```
HookyQR: This isn't required because we load the language configuration in the code. This would also require maintaining the same code in multiple locations. It's loaded in the code because we use RegExp for some of it and that's more than a pain to do this way.

language-configuration.json
```diff
@@ -0,0 +1,30 @@
+{
```
HookyQR: This file is not required, and increases maintenance costs.

ruby.js
```diff
@@ -42,8 +42,25 @@ const langConfig = {
...
+		["`", "`"]
+	],
+	surroundingPairs: [
+		["{", "}"],
+		["[", "]"],
+		["(", ")"],
+		["\"", "\""],
+		["'", "'"],
+		["`", "`"]
+	],
```
HookyQR: These are a new thing. Happy to have these included.

karuna: new syntax? let me check first

HookyQR: https://code.visualstudio.com/docs/extensionAPI/extension-points#_contributeslanguages

HookyQR:
@rebornix: Over to you for merge with expected version changes.

PR #48, opened 2016-08-25
Update Ruby Syntax

karuna:
Update the `ruby.tmLanguage` syntax using the latest from  https://github.com/textmate/ruby.tmbundle/blob/master/Syntaxes/Ruby.plist

- [x] Build pass!


HookyQR:
This is a pull directly from the original source of our current file (I remember it well).

Unfortunately, that doesn't work properly because the syntax highlighting for VS Code doesn't recognise `constant.language.symbol.hashkey.ruby` (and the like) but instead uses `constant.other.symbol.hashkey.ruby`. It's an annoying state, but it's true.

So, I vote no I'm afraid.

karuna:
I didn't know about this :(

HookyQR:
All good. There may be some updates in there, but a wholesale change would break the highlighting.

karuna:
@HookyQR do you know if there's a good vscode syntax highlighting engine documentation?

HookyQR:
Not really, but you can look at the classes that get set on certain things by opening the developer tools. That's how I found out they changed to `constant.other`. If you pull out one of the decent themes (Say `Material-theme` for example) it will show what classes get used in there too. The theme generator produces them (kind of), so you could be pretty confident that they're up to date.

rebornix:
@karuna both VS Code and Sublime are following TextMate's syntax parsing engine: https://manual.macromates.com/en/language_grammars and we have our own implementation https://github.com/Microsoft/vscode-textmate .

PR #50, opened 2016-09-01
Intellisense:  Allow specification of full paths.

scotthal:

- [X] Build pass!

Allow specification of full path to the Ruby interpreter and
rct-complete command for environments in which multiple versions
of Ruby and gems are installed.

### DIFF
--- a/ruby.js
+++ b/ruby.js
@@ -123,9 +123,12 @@ function balanceEvent(event) {
 }
 
 function completeCommand(args) {
+	let rctCompletePath = vscode.workspace.getConfiguration('ruby.rctComplete').commandPath;
+	args.push('--interpreter');
+	args.push(vscode.workspace.getConfiguration('ruby.interpreter').commandPath);
 	if (process.platform == 'win32')
-		return cp.spawn('cmd', ['/c', 'rct-complete'].concat(args));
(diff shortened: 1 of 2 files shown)


ruby.js
```diff
@@ -123,9 +123,12 @@ function balanceEvent(event) {
 }
 
 function completeCommand(args) {
+	let rctCompletePath = vscode.workspace.getConfiguration('ruby.rctComplete').commandPath;
+	args.push('--interpreter');
+	args.push(vscode.workspace.getConfiguration('ruby.interpreter').commandPath);
 	if (process.platform == 'win32')
```
rebornix: I suppose if we don't set a default value for this option, the value we get is an empty string, which makes this line of code not intuitive. Maybe pass in the default value?

scotthal: Done.

rebornix:
@scotthal thanks for your quick update! Love it!

danipindado:
how is this supposed to work for windows? 
i have tried
"ruby.interpreter.commandPath": "C:/Ruby24-x64/bin/ruby.exe"
"ruby.interpreter.commandPath": "C:\\Ruby24-x64\\bin\\ruby.exe"
and many more combinations. but in the end the ruby version present in PATH env. variable is executed.

PR #74, opened 2016-10-01
Add language configuration for erb

karuna:

- [x] Build pass!


language-configuration-erb.json
```diff
@@ -0,0 +1,28 @@
+{
+	"comments": {
+		"blockComment": [ "<%#", "%>" ]
```
HookyQR: Add regular HTML comments as well

language-configuration-erb.json
```diff
@@ -0,0 +1,28 @@
...
+		["(", ")"]
+	],
+	"autoClosingPairs": [
+		["{", "}"],
+		["[", "]"],
+		["(", ")"],
+		["`", "`"],
+		{ "open": "'", "close": "'", "notIn": ["string", "comment"] },
+		{ "open": "\"", "close": "\"", "notIn": ["string"] },
+		["<%", "%>"]
```
HookyQR: Add regular HTML element open/close

language-configuration-erb.json
```diff
@@ -0,0 +1,28 @@
...
+	],
+	"surroundingPairs": [
+		["{", "}"],
+		["[", "]"],
+		["(", ")"],
+		["\"", "\""],
+		["'", "'"],
+		["`", "`"],
+		["<%", "%>"]
```
HookyQR: Add regular HTML element open/close

language-configuration-erb.json
karuna: So it is better to do this?
```		["<", ">"],
		["%", "%"]
```

language-configuration-erb.json
karuna: So it is better to do this?

    ["<", ">"],
    ["%", "%"]

language-configuration-erb.json
karuna: So it is better to do this?

    ["<", ">"],
    ["%", "%"]

language-configuration-erb.json
karuna: I don't think we can do it

HookyQR: No, you're probably right. Perhaps leave the HTML comment and let the ruby language processing pickup the comment inside the erb tags.

HookyQR: No, you're probably right. Perhaps leave the HTML comment and let the ruby language processing pickup the comment inside the erb tags.

language-configuration-erb.json
```diff
@@ -0,0 +1,31 @@
...
+	],
+	"surroundingPairs": [
+		["{", "}"],
+		["[", "]"],
+		["(", ")"],
+		["\"", "\""],
+		["'", "'"],
+		["`", "`"],
+		["<", ">"],
+		["%", "%"]
```
HookyQR: Will this not work with:
```
		["<", ">"],
		["<%", "%>"]
```

karuna: It'll be `<%>`
It's either 

    ["<", ">"],
    { "open": "%", "close": "%", "notIn": ["string", "comment"] }

or

     ["<%", "%>"]


For HTML I think it's not going to be a problem since it's handled by emmet.

HookyQR: I think the last one (just the erb tags) is the best of those you've offered. I'll have a look and do some testing later.

HookyQR:
@rebornix This is good to merge + publish.
Thanks @karuna

HookyQR:
Hey @rebornix, still not seeing a new relase on the marketplace.

PR #98, opened 2016-12-03
Update sourceRoot in gulpfile.js.

### LINKED ISSUE #97: Minor issue - gulpfile.js specifies incorrect path for source maps.
*This is a minor issue, only affecting the ability to debug the debug adapter itself.*

`gulpfile.js` specifies a value of `../../src` for `sourcePath`.  However, the `out` directory is a sibling to the `src` directory, so the relative path should be just `../src`.

After I made this change in my fork, I was finally able to set breakpoints while debugging the extension.  Is there a good reason why the path is set the way it is?

seraku24:
Fixes #97.

---

- [x] Build pass!

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -55,7 +55,7 @@ gulp.task('internal-compile', function() {
 			// no inlined source
 			includeContent: inlineSource,
 			// Return relative source map root directories per file.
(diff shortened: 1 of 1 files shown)


rebornix:
@seraku24 thanks!

PR #108, opened 2017-01-02
Add common ERB snippets

jtokoph:
Adds `<% %>` and `<%= %>` to ERB snippets


rebornix:
@jtokoph thank you very much!

PR #127, opened 2017-03-11
Add support for custom ports to the Ruby debugger.

ukblewis:

- [] Build pass!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ before_install:
  - nvm install v5.10.0
  - nvm use v5.10.0
 before_script:
-  - npm install -g typescript
+  - npm install -g typescript@1.8.2
   - npm install
   - tsc -p ./src
 script: npm test
--- a/src/ruby.ts
+++ b/src/ruby.ts
@@ -195,6 +195,10 @@ export class RubyProcess extends EventEmitter {
                 runtimeArgs.push('-x');
             }
 
+            if (args.debuggerPort && args.debuggerPort !== '1234'){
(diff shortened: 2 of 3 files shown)


ukblewis:
Can I have some help with getting the tests to pass @rebornix ?

rebornix:
@ukblewis I'll take a look

ukblewis:
@rebornix Thank you - any progress so far?

src/ruby.ts
```diff
@@ -218,7 +222,7 @@ export class RubyProcess extends EventEmitter {
 
             this.debugprocess.stderr.on('data', (data: Buffer) => {
                 if (/^Fast Debugger/.test(data.toString())) {
-                    this.debugSocketClient.connect(1234);
+                    this.debugSocketClient.connect(args.debuggerPort);
```
rebornix: `args.debuggerPort` might be `undefined` so you may want to use `args.debuggerPort | 1234`, that's why the test cases fail.

.travis.yml
```diff
@@ -9,7 +9,7 @@ before_install:
  - nvm install v5.10.0
  - nvm use v5.10.0
 before_script:
-  - npm install -g typescript
+  - npm install -g typescript@1.8.2
```
rebornix: I already fix this in master, do you mind reverting this line of change.

rebornix:
LGTM, just two tiny comments and they should be real easy fix.

rebornix:
@ukblewis thanks, it's really awesome!

PR #128, opened 2017-03-21
Use new configurationDefaults

novak:
I noticed the indentation was not being set properly in the editor when opening up ruby files. This pull is rather small and could provide a starting point for other editor settings that would benefit from ruby specific defaults.

This adds a configurationDefaults entry to package.json to allow per
language editor settings by default.

- [] Build pass!

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,6 +77,11 @@
 			"language": "erb",
 			"path": "./snippets/erb.json"
 		}],
+		"configurationDefaults": {
+			"[ruby]": {
+				"editor.tabSize": 2
+			}
+		},
 		"configuration": {
 			"title": "ruby language settings",
 			"properties": {


rebornix:
I'm not sure if we should do this as forcing users to use `tabSize=2` as default is not that good to me.

novak:
I found it difficult to maintain the tab settings when creating new files in vs code, unless I'm missing something obvious there. In terms of defaulting to tabSize=2, most style guides and every editor I've used defaults ruby code to 2 spaces.

@rebornix do you find places where ruby isn't defaulted to tabSize=2?

MatthiasWinkelmann:
Would this override a top-level ```editor.tabSize``` in a user's configuration and require them to also add a ruby-specific line to his settings? 

Personally, I happen to use 3 spaces and I'd be quite annoyed if some extension thought it knew better than me.

(yes, I know... I'm just slightly to blind for two spaces, but don't feel like wasting horizontal space on 4)

novak:
@MatthiasWinkelmann it shouldn't override a top level setting as far as I know. User settings take precedence over extension settings.

PR #131, opened 2017-03-29
Add config to set Unicode handling on `ruby -wc` linting

danielgracia:
I use non-Ascii characters frequently, and since Ruby 2.0 source files are handled as UTF8 by default. `ruby -wc` however does not handle that correctly unless you pass a extra argument for forcing the character set (`-Ku`).

Because I don't know if people rely on the old behavior, this PR adds a flag for the RubyWC linter for enabling Unicode handling, disabled by default.

- [] Build pass!


rebornix:
@danielgracia thanks for your contribution, it looks good to me.

PR #145, opened 2017-05-03
Make stacktrace handling of missing files more robust

simonomis:
Improves the handling when files referenced in the stacktrace aren't available where vscode is running. Addresses two issues that were causing the debugger to throw an exception (and so fail to provide the call stack and current position) when debugging a rails app within a docker container:

- The rails app ran within the `/app` directory in the container, so files such as `/usr/local/lib/gems/railties-3.0.20/lib/rails/application.rb` weren't being filtered out by the check on line 259, which resulted in an exception when attempting to read the file (as the file only exists within the docker container)
- One stack frame reported by rdebug-ide had a line number of 0, which caused an exception when trying to get the line from the array (on line 275).

This PR changes the logic so that missing files aren't filtered out, because it's still useful to be able to see the stackframe even if you can't see the code. Instead, they are included as "filename:linenum", like this: `actionpack-3.0.20/lib/action_controller/base.rb:150 (Unknown Source:150)` (the gems directory path is stripped off to make it easier to read).

- [x] Build pass!

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -6,7 +6,7 @@
 
 import {DebugSession, InitializedEvent, TerminatedEvent, StoppedEvent, BreakpointEvent, OutputEvent, Thread, StackFrame, Scope, Source, Handles, Breakpoint} from 'vscode-debugadapter';
 import {DebugProtocol} from 'vscode-debugprotocol';
-import {readFileSync} from 'fs';
+import {readFileSync,existsSync} from 'fs';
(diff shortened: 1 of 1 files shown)


simonomis:
Let me know if you want some help testing this. I reckon it should be possible to do it locally without needing docker, but I haven't tried yet.

rebornix:
Thanks @simonomis it looks awesome!

PR #146, opened 2017-05-03
Stacktrace fails when file does not exists

Boudin:
I had an issue on an old ruby 1.9.3 project where the function generating the stack trace is failing because of a stack trace line being on file <internal:prelude>, which does not exists and a trace on line 0, which result in trying to access line -1 in the array containing files content.

The following pull request is a quick fix that : 
- does not try to fetch content from files that does not exists
- set source code to "<Source not available>" when the file or the line does not exists


Boudin:
Sorry, just saw the other pull request concerning the same issue. I'm closing this one for now.

PR #148, opened 2017-05-05
Fix the "debuggerPort" option

ukblewis:
So awkwardly, I've actually found that my code for debuggerPort didn't actually work - and now that I figured out how to nicely use the VSCode debug tools, I've fixed it :)

It's only a very small change.

The only thing that I have left on my hit list is figuring out how to debug processes that fork themselves into children (e.g. `puma` with `WEB_CONCURRENCY > 1`). But for now, this should actually make it possible to use this for my Ruby work :)

- [] Build pass!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ before_install:
  - nvm install v5.10.0
  - nvm use v5.10.0
 before_script:
-  - npm install -g typescript
+  - npm install -g typescript@1.8.2
   - npm install
   - tsc -p ./src
 script: npm test
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,10 +6,10 @@ rvm:
 
 before_install:
  - gem install ruby-debug-ide
- - nvm install v5.10.0
- - nvm use v5.10.0
+ - nvm install v6.9.1
+ - nvm use v6.9.1
 before_script:
-  - npm install -g typescript@1.8.2
+  - npm install -g typescript -v 1.8.2
   - npm install
   - tsc -p ./src
 script: npm test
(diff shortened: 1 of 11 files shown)


rebornix:
@ukblewis thanks for the turnaround!

ukblewis:
@rebornix Of course :) it's very nice to play with now and debug.

Thank you!

Do you have any ideas about the debugging processes that fork themselves into child processes problem? Right now this means that Rails won't debug out of the box on macOS and Linux since it defaults to a multi-process configuration (for performance and stability) but the child processes cannot be debugged properly. Having done some research, I believe that we need to configure the `--dispatcher-port` flag, but I don't really understand how it works. I'd love to hear any suggestions that you have of things to try! :)

rebornix:
@ukblewis I think it's related to rdebug-ide or debase we are using, which are written in Ruby and fast implementation of ruby debug. If they don't expose proper port for different processes, then we don't have a chance to listen to them, make sense? That's how far I've gone and as I'm not a professional ruby developer, I can't provide more insights but I'll definitely share more with you once I do.

senhalil:
Hi @rebornix I am trying to debug an application which uses `rake resque:workers`; however, per @benglewis 's https://github.com/rubyide/vscode-ruby/pull/148#issuecomment-299569882 I can follow the code till the fork (shown in the image)  and cannot enter the child processes. I can see the call stack of the main thread (2); however, cannot interact with the remaining threads (83-87).
vscode

Unfortunately the code I need to debug runs in the child processes and triggered when a task is submitted. Here is how my `launch.json` file looks like

```
        {
            "name": "Workers",
            "type": "Ruby",
            "request": "launch",
            "cwd": "${workspaceRoot}",
            "program": "/home/halil/.rbenv/versions/2.1.8/bin/rake",
            "useBundler": true,
            "env": {
                "COUNT": "5",
                "QUEUE": "*"
            },
            "args": ["resque:workers"]
        },
```
The moment I let the code continue (i.e., call the `thread.join()` on all threads) either with F5 or F11 (in fact F11 does nothing different since the debugger doesn't follow the code), one of the threads picks up the submitted task, runs the code I want to debug and prints out the result. 

I would really appreciate if you have an idea about how to debug such an application.

PR #152, opened 2017-05-10
Show method names in debugger call stack

simonomis:
Uses the ruby-method-locate module to get a list of the symbols (modules, classes, methods, etc) in each file mentioned in the stacktrace. Looks up the closest containing symbol for each line.

screen shot 2017-05-10 at 20 31 50

Format is: method name (with ::prefix for class/module methods), followed by full class/module name in brackets.

- [x] Build pass!

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -13,6 +13,7 @@ import * as childProcess from 'child_process';
 import * as path from 'path';
 import {DOMParser} from 'xmldom';
 import {RubyProcess} from './ruby';
+import {SymbolLocator} from './symbols';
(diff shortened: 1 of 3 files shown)


timriley:
This looks great! @simonomis, I wonder if you considered using prefixes for both types of methods (e.g. `#i_am_instance_method` and `.i_am_class_or_module_method`) - this is consistent with Ruby documentation and might make that method list look more consistent?

simonomis:
Thanks! That's a much better idea @timriley, I'll try it out...

simonomis:
I've pushed up the change to use the new method prefixes and updated the screenshot.

PR #182, opened 2017-07-19
Fix type of ruby linter property

5t111111:
In `ruby.lint` settings, the `ruby` linter can have not only `boolean` but also `object` value, for example, the readme says:

```js
"ruby.lint": {
	"ruby": {
		"unicode": true //Runs ruby -wc -Ku
	},
```

However, since current property of `ruby` linter only accepts `boolean`, setting an object as `ruby` linter configuration complains that there exists a type mismatch.

So I have changed the property setting of `ruby` linter to accept either `boolean` or `object`.

- [] Build pass!

### DIFF
--- a/package.json
+++ b/package.json
@@ -117,7 +117,7 @@
 					"description": "Set individual ruby linters to use",
 					"properties": {
 						"ruby": {
-							"type": "boolean",
+							"type": ["boolean", "object"],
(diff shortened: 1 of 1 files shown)


rebornix:
@5t111111 thanks for your help on this!

Issue #203, opened 2017-09-07
Extra indentation of def when prefixed with "private"


faustinoaq:
### Your environment

- `vscode-ruby` version:  0.15.0
- Ruby version: ruby 2.4.1p111 (2017-03-22 revision 58053) [x86_64-linux]
- VS Code version: 1.15.0
- Operating System: Manjaro Linux KDE Gellivara
- Hardware (optional): Intel Celeron late 2014 

### Expected behavior

When I type

```ruby
class ClassName
  private def miprivatefunc(data)
  end
end
```
The end of `miprivatefunc` should be indented correctly when `autoIndent` is enabled.

### Actual behavior

![Imgur](https://i.imgur.com/wY66iNe.gif)

![Imgur](https://i.imgur.com/bVYcuZo.gif)

### Steps to reproduce the problem

1. Open a ruby file.
2. Ensure `autoIndent` is enabled.
3. Write some function with indentation (by example, inside a class), then try to write end keyword

As workaround I disabled `autoIndent` in `settings.json`

```json
{
  "editor.autoIndent": false
}
```

This issue also happens with vscode-crystal-lang :point_right: https://github.com/crystal-lang-tools/vscode-crystal-lang/issues/18

faustinoaq:
Already fixed on [Crystal](https://github.com/crystal-lang-tools/vscode-crystal-lang/commit/a9105e926485d4e8d8bc973bf246f7ffc23dc9ba) and default [Ruby](https://github.com/Microsoft/vscode/pull/34795) extension :sweat_smile:

PR #206, opened 2017-09-15
Fix bug in indentation of "private def" methods

juicemia:
There was a bug in the plugin's auto-indentation where a method declared
as private on the same line as its definition would fail to indent
correctly.

"private def" was added as a keyword in the regex to keep the regex
simple and explicit. Reading the code shows you that "private def" is
something for which the editor should increase the indentation level
automatically.

Fixes rubyide/vscode-ruby#203

- [] Build pass!

### DIFF
--- a/src/ruby.ts
+++ b/src/ruby.ts
@@ -15,7 +15,7 @@ export function activate(context: ExtensionContext) {
 	// register language config
 	vscode.languages.setLanguageConfiguration('ruby', {
 		indentationRules: {
-			increaseIndentPattern: /^\s*((begin|class|def|else|elsif|ensure|for|if|module|rescue|unless|until|when|while)|(.*\sdo\b))\b[^\{;]*$/,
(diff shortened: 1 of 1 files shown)


faustinoaq:
Hi, what's about `when` statement ? :sweat_smile: 

when

faustinoaq:
@juicemia take a look https://github.com/Microsoft/vscode/pull/34795

private-when

juicemia:
@faustinoaq so it looks like they fixed it then? Can this PR be closed?

faustinoaq:
@juicemia I think this PR is still valid because is fixed on `Microsoft/vscode` but not on `rubyide`. So, this extension can replace default settings and show indentation issue again.

faustinoaq:
Fixed by #231

wingrunr21:
Looks like this was superseded by another PR so I'm going to close it.

PR #223, opened 2017-11-21
improve indentation

wyattisimo:
* add private/protected keywords to increaseIndentPattern
* remove when keyword from decreaseIndentPattern

Fixes https://github.com/rubyide/vscode-ruby/issues/203

Make sure these boxes are checked before submitting your PR -- thanks in advance!

- [ ] Build pass!
- [x] Follow VS Code [Coding Guidelines](https://github.com/Microsoft/vscode/wiki/Coding-Guidelines)

PR #231, opened 2017-12-12
Update indent pattern

TeeSeal:
- [ ] Build pass!

Refer the following pull request and the issues linked in it:
https://github.com/Microsoft/vscode/pull/40049

@rebornix 
This PR seems to come into a bit of conflict with #223, should make a decision on which one we're going to use.
 
Here's the thing:
1. If `case` doesn't indent the next line and `when` doesn't unindent, the indentation will be ok only while we're adding `when` clauses, but will break when we use `else` and/or `end` (because these 2  do unindent). The result of using the changes made in #233 would be this:
```rb
  case thing
  when 2 then true
  when 4 then false
else nil
end
```

2. With this PR the `case` statement does indent the next line but any of `when`, `else` and `end` will unindent leading to a desired output:

```rb
  case thing
  when 2 then true
  when 4 then false
  else nil
  end
```

### DIFF
--- a/src/ruby.ts
+++ b/src/ruby.ts
@@ -15,7 +15,7 @@ export function activate(context: ExtensionContext) {
 	// register language config
 	vscode.languages.setLanguageConfiguration('ruby', {
 		indentationRules: {
-			increaseIndentPattern: /^\s*((begin|class|((private|protected)\s+)?def|else|elsif|ensure|for|if|module|rescue|unless|until|when|while|case)|([^#]*\sdo\b)|([^#]*=\s*(case|if|unless)))\b([^\{;]|("|'|\/).+\4)*$/,
+			increaseIndentPattern: /^\s*((begin|class|((private|protected)\s+)?def|else|elsif|ensure|for|if|module|rescue|unless|until|when|while|case)|([^#]*\sdo\b)|([^#]*=\s*(case|if|unless)))\b([^\{;]|("|'|\/).*\4)*$/,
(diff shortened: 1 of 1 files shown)


TeeSeal:
I also made the changes for `private` and `protected` defs the same way they are in the built in extension for Ruby of VS Code. This can now be considered in favor of #223 and #206.

I also edited the `[^\{;]` part to just `[^;]`. I didn't see the point of that curly brace there, and it made things such as:
```
  if string = /\d{4}/
  #not indent properly
end
```

wingrunr21:
ping @rebornix

wingrunr21:
Hi @TeeSeal 

Would you be willing to reopen this PR against my fork?

Fork: https://github.com/wingrunr21/vscode-ruby-lang

See https://github.com/rubyide/vscode-ruby/issues/239#issuecomment-359279950 for more information

Thanks!

TeeSeal:
@wingrunr21 okay, will do!

rebornix:
@TeeSeal your PR LGTM, thanks for your contribution.

wingrunr21:
@rebornix 

@TeeSeal made additional updates to that intent and unindent patterns in the PR to my fork. It fixes a few other issues that are still in the patterns in this PR.

wingrunr21/vscode-ruby-lang#9

TeeSeal:
Ye I thought this place is dead so I didn’t bother updating this source branch

TeeSeal:
There are still some issues they are a bit harder to handle, Ill investigate those, see if I can do anything about them and open another PR

rebornix:
@TeeSeal thanks, look forward to it ;)

PR #248, opened 2018-01-24
Use Solargraph for Code Completion

castwide:
Replace rcodetools code completion with a dependency on the castwide.solargraph extension.


rebornix:
@castwide thanks for your contribution. Your extension is already super great, thanks for brining this feature here.

Can you have an option and then users can choose which one to use, rcodetools or solargraph?

castwide:
Thanks. I'll look into making an option setting for the completion method.

rebornix:
@castwide I like the PR is backwards-compatible. As it's a breaking change, I'll suggest we ship it alone later instead of having together with all other features/bug fixes that I didn't shipt for half a year, how does that sound?

castwide:
Sounds good to me. I'll release the new version of vscode-solargraph in the next couple days. It'll start using the `ruby.codeCompletion` setting whenever that version of vscode-ruby gets released. In the meantime it'll work the same as before, so no hurry.

rebornix:
Let's get ready for next release ;)

aivicevic:
When is this set to be released? Really want to be able to navigate to definitions... this is painful.

wingrunr21:
#259 would need to be merged first I think

castwide:
@aivicevic If you install the [castwide.solargraph](https://marketplace.visualstudio.com/items?itemName=castwide.solargraph) extension, it should work as is. This PR and #259 just provide optional settings to avoid running redundant providers. You might see duplicate entries, for example, when both extensions detect the same method definition.

Note that go to definition is a recent addition to Solargraph. If you run into any bugs, please consider submitting an issue to the [repo](https://github.com/castwide/vscode-solargraph).

aivicevic:
@castwide I have the extension installed and it works for some things. It seems if I try to "Go to Definition" on anything from within a file that's under the "./spec/" directory, it will not find anything. If I try from outside that directory, it works as well as expected. I have tried customizing the .solargraph.yml file and adding it to my workspace root but no luck. Any suggestions? If you'd prefer I open an issue directly on that repo I can.

castwide:
@aivicevic ./spec and ./test directories are excluded from code maps by default. You can use `solargraph config` to generate a default configuration file, which will provide the settings Solargraph would use in the file's absence. You should be able to override the default by setting `exclude` to an empty array:

```
exclude: []
```

Including specs in maps has some potentially undesirable side effects, such as including test symbols in completion items when you're working on application code. (Specs used to have a detrimental impact on performance as well, although that problem may be better mitigated now.) I'm working on ways for Solargraph to be more aware of a file's expected environment.

aivicevic:
@castwide That's what I've already done, unfortunately I am still unable to navigate to definitions to a large number of methods and/or variables. Not sure if I'm doing something wrong or if it is an issue.

Currently, I am using VSCode v1.20.0, Ruby extension v0.16.0, and Solargraph extension v0.12.0 + solargraph gem v0.17.1. I have a solargraph.yml config file in the root of my workspace with an empty array for "exclude" as you mentioned (just to test for now) and I am still unable to navigate to definitions for anything within my ./spec folder.

PR #258, opened 2018-01-31
Added .arb to file extensions list

### LINKED ISSUE #256: .arb file extension support
[Arbre](https://github.com/activeadmin/arbre) is a ruby gem that allows you to build view components using plain ruby (instead of embedded ruby, as in erb). In Arbre, you can build partials and views just like with ERB, but the file extension is different (`.arb` instead of `.erb`).

I think supporting this file extension should be fairly straightforward. The code inside the file is just ruby (no fancy syntax), so it should be as easy as recognizing the file extension, and using the right language for it.

In the meantime, an easy workaround for anyone interested is to add this to your settings config:

```js
  "files.associations": {
    "*.arb": "ruby"
  }
```

dkniffin:
Fixes #256

- [x] Build pass!

### DIFF
--- a/package.json
+++ b/package.json
@@ -266,6 +266,7 @@
         ],
         "firstLine": "^#!\\s*/.*(?:ruby|rbx|rake)\\b",
         "extensions": [
+          ".arb",
           ".builder",
           ".cgi",
           ".fcgi",


rebornix:
@dkniffin thanks for your help!

PR #259, opened 2018-02-04
Add `ruby.intellisense` configuration option

castwide:
- [x] Build pass!

This PR adds the `ruby.intellisense` configuration to allow method selection for go to/peek definition similar to `ruby.codeCompletion`. Valid options are `solargraph`, `rubyLocate`, and `none`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -139,6 +139,12 @@
           "default": "solargraph",
           "description": "Method to use for code completion."
         },
+        "ruby.intellisense": {
+          "type": "string",
+          "enum": ["solargraph", "rubyLocate", "none"],
+          "default": "solargraph",
+          "description": "Method to use for intellisense (go to definition, etc.)."
+        },
         "ruby.rctComplete.commandPath": {
           "type": "string",
           "default": "rct-complete",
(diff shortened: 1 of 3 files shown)


wingrunr21:
Should this PR also remove the [extension dependency on solargraph](https://github.com/rubyide/vscode-ruby/blob/master/package.json#L72-L74)?

castwide:
The extension dependency is still necessary for the `solargraph` options to work. This PR helps ensure that only one intellisense provider is registered, so "go to definition" doesn't return duplicated entries.

wingrunr21:
Gotcha. I guess my concern is if users select `rubyLocate` or `none`, they are going to wonder why the solargraph extension is a requirement if they are not using it.

castwide:
Fair point. If neither `ruby.codeCompletion` nor `ruby.intellisense` is set to `solargraph`, the extension is not strictly required (although it will still provide access to documentation through the command palette). Making it a dependency is just the easiest way to integrate it.

wingrunr21:
Ok. I'm going to open an issue on the topic (as I'm not convinced having it as a hard dependency is the right choice) but I also don't want to hold up this change. A standalone issue will allow other commentary and we can always remove the dependency later.

Issue #252, opened 2018-01-26
Linters and vim mode(s) don't get along


onyxraven:
### Your environment

- `vscode-ruby` version: 0.16.0 (and 0.15.0 it turns out)
- Ruby version: 2.4.3
- VS Code version: Version 1.19.3 (1.19.3)
- Operating System: OSX 10.13.3 (17D47)
- Hardware (optional): MacBook Pro (Retina, 15-inch, Mid 2014)

Make sure you have `ruby`, `ruby-debug-ide` and `ruby-debug-basex19` installed before submitting your issue -- thank you !

### Expected behavior

When using Vim or amVim extensions along with Ruby linters enabled, typing should not be delayed every character for lints to complete.

### Actual behavior

It seems like for every linter enabled, they are running on the main input thread and blocking the inputs.  This looks to even get in between the keypress and the autocomplete/suggestions.

### Steps to reproduce the problem

It seems like using either of those extensions, along with any enabled linters (which arguably do hook all editor inputs for modes) cause the issue to show very obviously.  Editing other file types with the extensions is not a problem.  Disabling all linters makes inputs work again without delay.  

----------

I'm happy to help with more investigation, but I just need some guidance on where to look next.

onyxraven:
I just downgraded to 0.15 (using the helpful link [in a different issue](https://github.com/rubyide/vscode-ruby/issues/251#issuecomment-360806583)) and it seems to be still happening in the same way. So I'm thinking it has something to do with the 1.19.3 update.  I'll check the issue board for the main VSCode project to see if there is anything related.

castwide:
I don't use Vim or amVim, but I had a similar problem. I got the best performance (short of no linters at all) using RuboCop alone with the latest gem.

MythicalFish:
Same issue here, can confirm that disabling the linters solves the lag.

wingrunr21:
Can some of you post your linter configs?

I also don't see this issue using VIM + Rubocop. I'm wondering if the changes made in #232 might be affecting something there (although linter performance wasn't great before that either).

cyangOL:
I have the same issue. Typing is noticeably slow when vim and ruby are the only enabled plugin.

onyxraven:
in 1.6.0 I was able to reproduce with the following config

```javascript
    "ruby.lint": {
        "rubocop": false,
        "ruby": true, //Runs ruby -wc
        "fasterer": false,
        "debride": false,
        "ruby-lint": false
    },
```

[...]

PR #264, opened 2018-02-11
Debounce linting to prevent running on every keypress

Sam-Killgallon:
I've been having this issue https://github.com/rubyide/vscode-ruby/issues/252 where the linters run after every keypress and cause significant input lag.

This PR makes the linters only run when typing has finished, instead of after each letter.

Before:
slow_linting

After:
debounced_linting

EDIT: Updated to include a configuration option for the delay.

### DIFF
--- a/src/ruby.ts
+++ b/src/ruby.ts
@@ -10,6 +10,7 @@ import { RubyDocumentFormattingEditProvider } from './format/rubyFormat';
 import * as utils from './utils';
 import { registerTaskProvider } from './task/rake';
 import { Config as LintConfig } from './lint/lintConfig';
+import * as debounce from 'lodash/debounce';
 
 export function activate(context: ExtensionContext) {
 	const subs = context.subscriptions;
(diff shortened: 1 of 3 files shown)


src/ruby.ts
```diff
@@ -120,7 +121,8 @@ function registerLinters(ctx: ExtensionContext) {
 	}
 
 	ctx.subscriptions.push(vscode.window.onDidChangeActiveTextEditor(executeLinting));
-	ctx.subscriptions.push(vscode.workspace.onDidChangeTextDocument(executeLinting));
+	// Debounce linting to prevent running on every keypress, only run when typing has stopped
+	ctx.subscriptions.push(vscode.workspace.onDidChangeTextDocument(debounce(executeLinting, 500)));
```
onyxraven: You might make the debounce delay configurable?  Some of the other delays (eg autocomplete recommendations) are configurable.

Sam-Killgallon: I hardcoded it as I didn't think there would be too much need to change it, and didn't want to needlessly add extra config options.

If it would be useful for some to be able to configure it though, I'm happy to do so.

onyxraven: Ultimately up to maintainers, but any time I see a delay I like it to be configurable (unit testing, different user preferences, etc).

papa-whisky:
@rebornix or @wingrunr21 is there any chance of getting this merged? This and the broken syntax highlighting has made writing ruby in VSCode pretty painful for the past month or so ☹️ 

Appreciate all your efforts on the extension!

wingrunr21:
I think it needs a config option

doudou:
@rebornix, @wingrunr21 the OP added the configuration option ... could we get this merged now ? I can't use rubocop at all without it, and it's getting hard to track all these PRs I add on top of my local vscode-ruby package...

wingrunr21:
I can merge it but @rebornix can you push a release?

onyxraven:
This fixed my issues and made this plugin completely great again (especially with the solargraph additions) - any chance we can get an official release so I don't have to run it from the 'master' branch locally?  @rebornix :)

rebornix:
I’ll do a release today.

onyxraven:
@rebornix It looks like the release didn't happen last week. Just a friendly reminder.  Thanks for your work here!

rebornix:
Didn't make that happen as promised, sorry for that. 0.18.0 is released now.

PR #268, opened 2018-02-19
add settings to control bundler usage for linters

doudou:
The pathToBundler and useBundler settings allow to control whether
and how linters will be using bundler. Not setting useBundler retains
the current auto-detection behavior, but setting it to either true
or false will avoid the call to bundler altogether.

- [x] Build pass!


wingrunr21:
Awesome, thanks!

PR #279, opened 2018-03-03
Modify upstream grammar to support optimistic syntax. Resolves #251.

### LINKED ISSUE #251: Some, but not all, syntax highlighting has stopped working.
### Your environment

- `vscode-ruby` version: 0.16.0
- Ruby version: 2.4.3
- VS Code version: 1.19.3
- Operating System: macOS Sierra (10.12.6)
- Hardware (optional): MacBook Pro (15-inch, 2016)

#### Other Ruby VSCode Plugins:
- ruby-rubocop
- rspec-snippets

### Expected behavior
Syntax highlighting should work. Method names should be highlighted in yellow, symbols in blue, etc.

### Actual behavior
Method name highlighting, variable name highlighting, and symbol highlighting have stopped working. There may be more, but these are the most obvious culprits right now. I'll keep exploring to see if there's more.

Highlighting for things like require, module/class declaration and names, def/end, if/else/end, and a lot of other things are still working normally.

Here's a screenshot showing the state of some functionality:
screen shot 2018-01-25 at 3 05 16 pm

### Steps to reproduce the problem
This happened without any intervention on my part at some point today. I'm assuming the 0.16.0 update did this because it was the first update in 6 months and was released today.

wingrunr21:
The original syntax highlighting was based on an "optimistic" syntax highlighting grammar. This change allows us to continue to use the upstream Atom grammar but will insert the correct grammar scopes to still support the optimistic highlighting.

See [here](https://github.com/rubyide/vscode-ruby/issues/251#issuecomment-363093172) for information on what these scopes mean and why.

Note that I could not discover a way to support `constant.other.symbol` and `constant.language.symbol` at the same time. This means Ruby symbols will be highlighted the same as core language constants (like `nil`). In reality this is probably ok as most themes with explicit support appear to map the same color to both types.

Before:
syntax_highlighting_before

After:
syntax_highlighting_after

### DIFF
--- a/scripts/meta.function-call.ruby.json
+++ b/scripts/meta.function-call.ruby.json
@@ -1,0 +1,14 @@
+{
+  "begin": "(?=[a-zA-Z0-9_!?]+\\()",
+  "end": "(?<=\\))",
+  "name": "meta.function-call.ruby",
+  "patterns": [
+    { "include": "#nest_function_parens" },
+    { "include": "#known_function_names" },
+    {
+      "match": "([a-zA-Z0-9_!?]+)(?=\\()",
+      "name": "entity.name.function.ruby"
+    },
+    { "include": "$self" }
+  ]
+}
--- a/scripts/update-grammar.js
+++ b/scripts/update-grammar.js
@@ -112,7 +112,6 @@ function getGrammarModifier(grammarName) {
 	switch(grammarName) {
 		case 'ruby':
 			return modifyRubyGrammar;
-			break;
 	}
 }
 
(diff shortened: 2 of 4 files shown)


roflomglol:
Hello,

After march update I've noticed odd highlighting and found out in the changelog that there were changes regarding this. So the question is, in this particular example, is method call `.three['a']` highlighted as expected?

image

Thanks

wingrunr21:
@roflomglol would you mind moving this to a new issue? I can follow up there

roflomglol:
Yeah, sure. Sorry, thats probably not a good place to discuss it. :) Here is the [issue](https://github.com/rubyide/vscode-ruby/issues/283)

PR #287, opened 2018-03-07
add --force-exclusion flag to reek options

pelted:
Force excluding files specified in the configuration `exclude_paths` even if they are explicitly passed as arguments.

Resolves problem with exclude_paths in the `.reek` config being ignored since we run reek on single files rather than the whole project.

Example that now works after this PR.

```yaml
exclude_paths:
  - db/migrate
```

- [x] Build pass!


matkoniecz:
@rubyide Can you consider reviewing/merging/rejecting this PR? I though about submitting one, but PR older than year still waiting for reaction are a bit discouraging.

PR #290, opened 2018-03-12
Fix typo

msysyamamoto:

- [x] Build pass!

### DIFF
--- a/syntaxes/ruby.cson.json
+++ b/syntaxes/ruby.cson.json
@@ -401,7 +401,7 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


wingrunr21:
Unfortunately merging this will be undone as soon as we pull in new updates from the Atom Ruby grammar. See [here](https://github.com/atom/language-ruby/blob/master/grammars/ruby.cson#L425) where they have a typo there as well. If you can open a PR against that repo to fix it then we should get it pulled in here once they merge.

Thanks!

PR #298, opened 2018-03-13
Added support for ruby.format, defaults to false (no formatting).

### LINKED ISSUE #295: Support Rufo for formatting
Add support for using [rufo](https://github.com/ruby-formatter/rufo) for formatting instead of Rubocop

gurgeous:
Added support for ruby.format, defaults to false (no formatting). Set it to 'rubocop' to enable rubocop formatting. Fixes #295.

### DIFF
--- a/package.json
+++ b/package.json
@@ -284,6 +284,12 @@
               }
             }
           }
+        },
+        "ruby.format": {
+          "type": [ "boolean", "string" ],
+          "enum": [ false, "rubocop" ],
+          "default": false,
+          "description": "Which system to use for formatting, or false for no formatting"
         }
       }
     },
--- a/readme.md
+++ b/readme.md
@@ -136,7 +136,19 @@ Settings available (in your VSCode workspace) for each of the linters:
 ```
 ## Formatting
 
+The VS Code Ruby extension can automatically format your Ruby files whenever you save.
+
+### Rubocop
+
+Set `ruby.format` to `rubocop` to enable rubocop formatting on save.
+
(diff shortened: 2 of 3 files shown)


src/format/RuboCop.ts
```diff
@@ -1,82 +1,180 @@
 "use strict";
-import * as fs from 'fs';
-import * as path from 'path';
-import * as cp from 'child_process';
-import * as os from 'os';
 
+import * as cp from "child_process";
```
wingrunr21: Whitespace changes here. [VSCode's guidelines use tabs](https://github.com/Microsoft/vscode/wiki/Coding-Guidelines)

gurgeous: I'd be happy to dig up a `prettier` config that matches the VSCode guidelines and start using it for vscode-ruby... Might result in a lot of changes, though. Especially if we decided to just run it in one PR. Separate PR, perhaps?

wingrunr21: Converting the whole codebase should be a standalone PR. In the mean time...hand format? Sorry but following the VSCode guidelines is in the PR template

gurgeous: See the new commit in #297. If that one looks good, let me know and I'll make the same change here.

I re-ran prettier with the following options. We should start using this ASAP and then we could avoid these discussions forever. :)

```
{
  "printWidth": 100,
  "singleQuote": true,
  "trailingComma": "es5",
  "useTabs": true
}
```

wingrunr21: Good to know on the prettier options. We can add that at a later time. In the mean time, people will have to continue to suffer through the code style guidelines by hand.

PR #309, opened 2018-03-28
Remove solargraph dependency and update codeCompletion and intellisense options

castwide:
* The solargraph dependency is removed.
* `ruby.codeCompletion` and `ruby.intellisense` use `false` to disable rcodetools and rubyLocate respectively.
* When rubyLocate is disabled, the `ruby.reloadProject` command displays informative messages.

### DIFF
--- a/package.json
+++ b/package.json
@@ -131,16 +131,16 @@
           "isExecutable": true
         },
         "ruby.codeCompletion": {
-          "type": "string",
-          "enum": ["rcodetools", "other"],
-          "default": "rcodetools",
-          "description": "Method to use for code completion. Use `other` if another extension provides this feature."
+          "type": ["boolean", "string"],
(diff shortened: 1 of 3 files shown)


wingrunr21:
Thanks! I'll merge this once CI passes

hstove:
Thanks for this change! Unfortunately I had to disable this extension because solargraph was using a lot of memory. Any plans for a version bump with this included?

castwide:
@hstove Although it's still in the dependencies, the latest version of the solargraph extension should not use any resources when you disable it in the user settings. The previous version would still run a process in the background; the current version does not.

If you want to try it, make sure that both `ruby.intellisense` and `ruby.codeCompletion` are set to a value other than `solargraph` (`none` is a valid option). Let me know if it still gives you trouble.

PR #311, opened 2018-03-31
Code Cleanup - first pass

wingrunr21:
First pass at cleaning up the code. This pass just basically moves most of the providers into their own files for better organization. `ruby.ts` got a bit of cleanup via the linter/prettier but nothing substantial.


gurgeous:
This looks great. Gotta get formatting turned on, though. Perhaps some tests would help us feel better about changes like this? I wrote some for my other extension and found them quite handy.

PR #326, opened 2018-04-10
[Discussion] [Live Share] Restricting language services to local files

lostintangent:
In preparation for [Visual Studio Live Share](https://aka.ms/vsls) adding support for "guests" to receive remote language services for Ruby, this PR makes the following changes:

1. Updates the current `DocumentSelector` to be limited to `file` and `untitled` (unsaved) files.
1. Prevents the linter from running on workspaces that don't appear to have a root path (which a Live Share/virtual file system wouldn't). The task and intellisense providers already did this, so I made some minor refactorings to gate them all behind the same check.

With these changes, when someone has the Ruby extension installed, and joins a Live Share session (where files use the `vsls:` scheme), their language services will be entirely derived from the remote/host side, which provides a more accurate and project-wide experience (guests in Live Share don't have local file access to the project they're collaborating with).

If someone joins a Ruby project using Live Share, and doesn't have the Ruby extension installed, then they will automatically receive language services from the host (which is awesome! 🎉), so this PR is simply an optimization for the case where collaborating developers both have the Ruby extension installed. Additionally, this wouldn't impact the "local" Ruby development experience.

*Note: As an example, the TypeScript/JavaScript language services that come in-box with VS Code [already have this scheme restriction](https://github.com/Microsoft/vscode/blob/master/extensions/typescript-language-features/src/utils/fileSchemes.ts#L12), and so this PR replicates that behavior.*

### DIFF
--- a/src/format/rubyFormat.ts
+++ b/src/format/rubyFormat.ts
@@ -6,7 +6,7 @@ import { AutoCorrect } from './RuboCop';
 export class RubyDocumentFormattingEditProvider implements vscode.DocumentFormattingEditProvider {
 	private autoCorrect: AutoCorrect;
 
-	public register(ctx: vscode.ExtensionContext) {
+	public register(ctx: vscode.ExtensionContext, documentSelector: vscode.DocumentSelector) {
 		// only attempt to format if ruby.format is set to rubocop
 		if (vscode.workspace.getConfiguration("ruby").get("format") !== "rubocop") {
 			return;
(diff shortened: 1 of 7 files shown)


lostintangent:
@wingrunr21 The failing tests seemed to be related to a debug adapter issue (aside from this PR), but that appears to have been resolved in master now.

wingrunr21:
Awesome, thanks for this.

@gurgeous FYI on this feature set for your implementation

lostintangent:
@wingrunr21 Thanks! Out of curiosity: do you have any plans to update the extension in the marketplace anytime soon?

PR #333, opened 2018-04-17
Detach in Remote Mode

dlanileonardo:
In Remote Mode (Attach) Exit and Restart Button send Quit command instead of Detach and this causes stop Rails Server.

Is good to Work with Docker (Compose).

Look this PR - https://github.com/ruby-debug/ruby-debug-ide/pull/145

- [x] Build pass!


stefansedich:
I think this behavior should be configurable as the current behavior is what I prefer to happen, and a change would cause issues with the experience I have for our debugging environment.

wingrunr21:
Given the feedback here + the #394 being opened I'm closing this PR. Thanks!

PR #366, opened 2018-07-04
Language Server - proof of concept

wingrunr21:
This is a proof of concept internal language server. The server itself is opt-in and will gradually replace features in the core extension. This means if you don't like the server's features you can turn it off and go back to the current feature set.

This may not look like much but `tree-sitter` was a bit of a moving target. I also did a ton of research on how to implement some of the other features/problems that are currently in the extension.

**Features**
* Parses current document on server into an AST using `node-tree-sitter`. This is the same experimental syntax highlighter Atom is using. This positions us quite well to support things like semantic syntax highlighting!
* Balanced pair highlighting - The highlight provider will only send back highlights for balanced pairs. This is inline with the current highlighter (except this one works correctly). Next steps will be expanding the provider to highlight other language constructs (like all occurrences of methods or variables within a document)
* *new feature!* Syntactic code folding - I implemented the [proposed syntactic folding provider](https://github.com/Microsoft/vscode-languageserver-protocol-foldingprovider) to allow correctly identified region folding. Next steps will be to implement the other two types of supported folds (multiline/block comments and requires).

**Merge Blockers**
* Update documentation
* Figure out best way to distribute `node-tree-sitter` and `tree-sitter-ruby` against the version of Electron VSCode is running on


wingrunr21:
I'm currently poking at the `tree-sitter`/`tree-sitter-ruby` problem via `prebuild` and its supportive tooling. Hopefully that's a quick fix.

wingrunr21:
Tree sitter related PRs landed this weekend. Once new releases are cut with prebuilt binaries I'll update this branch and prep a release!

numbcoder:
Nice job! 
Can't wait for this PR!

server/src/providers/ASTProvider.ts
```diff
@@ -0,0 +1,93 @@
...
+					tree.edit({
+						startIndex, // index in old doc the change started
+						oldEndIndex, // end index for old version of text
+						newEndIndex: range.end.line * range.end.character - 1, // end index for new version of text
+
+						startPosition: Position.FROM_VS_POSITION(range.start).toTSPosition(), // position in old doc change started
+						oldEndPosition: this.computeEndPosition(startIndex, oldEndIndex, tree), // position in old doc change ended.
+						newEndPosition: Position.FROM_VS_POSITION(range.end).toTSPosition(), // position in new doc change ended
+					});
+					tree = this.parser.parse(text, tree);
```
maxbrunsfeld: I think that you could move this `parse` call out of the for loop and just parse once with the final text.

wingrunr21: Ya that’s on my refactor list. I’m planning on building out a more robust object tree on my second pass here. Also need to figure out how to support incremental updates with the API VSCode exposes. I don’t think this is quite right

wingrunr21:
Planning to merge this and cut a release this weekend

Issue #261, opened 2018-02-09
Incorrect boolean flag to turn on rubocop


thawatchai:
### Your environment

- `vscode-ruby` version: 0.16.0
- Ruby version: 2.5.0
- VS Code version: 1.20.0
- Operating System: macOS 10.13.3
- Hardware (optional): 

Make sure you have `ruby`, `ruby-debug-ide` and `ruby-debug-basex19` installed before submitting your issue -- thank you !

### Expected behavior
This should turn on the linter and rails support.
```
    "ruby.lint": {
        "rubocop": {
            "lint": true,
            "rails": true
        }
    }
```

### Actual behavior
This turns on the linter and rails support.
```
    "ruby.lint": {
        "rubocop": {
            "lint": false,
            "rails": false
        }
    }
```

### Steps to reproduce the problem
Put the config in my user settings and change from `true` to `false`.

thawatchai:
This is probably related to #229.

rpellerin:
Any updates on this?

andrewmcodes:
Is this still an issue with 0.20.0?

happysean-lavabit-com:
@andrewmcodes yes this is still an issue

- vscode-ruby 0.21.0
-ruby version: 2.5.3p105
-vs code version 1.30.0
-operating system: windows 10 enterprise 1709 build 16299.846

wingrunr21:
This should be working correctly on `v0.22.2` and up with language server based linting

PR #388, opened 2018-09-18
fix docs to reflect how to lint with rubocop

andrewmcodes:
After running through the issues and spending a few hours trying to figure out why rubocop, I came to the following conclusions:

Rubocop linting works if the settings are as follows: 
```
	"rubocop": {
		"only": ["SpaceInsideBlockBraces", "LeadingCommentSpace"],
		"lint": false,
		"rails": false
	},
```
or
```
	"rubocop": {
		"rails": true
	},
```
I am opening this PR so that others can give input, as I could not find the core issue on my own. But, it looks like if `lint: true` is not included in the rubocop settings, linting works as expected. This is consistent the behavior described in #261.

- [] The build passes
- [x] TSLint is mostly happy
- [x] Prettier has been run

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,6 +154,7 @@ Settings available (in your VSCode workspace) for each of the linters:
 	"rails": true //Run extra rails cops
 }
 ```
+
 ## Formatting
 
 The VS Code Ruby extension can automatically format your Ruby files whenever you save.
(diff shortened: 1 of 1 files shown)


javiervidal:
Yes, `lint: true`does not work inside `rubocop`.

wingrunr21:
This shouldn't be needed anymore with #405 merged. It should make the RuboCop linting much more resilient.

Issue #125, opened 2017-03-09
Feature request: Attaching debugger to existing process


pelletencate:
The latest (beta) versions of `ruby-debug-ide` have support for attaching a Ruby debugger to existing processes. A nice implementation is done in Rubymine, where you can get a list of currently running ruby processes and select one to attach a debugger to. It works flawless on OS X, whether I attach to a rails server, Sidekiq, even a Zeus slave, and the nice thing is, killing the debugger detaches it again, making the original process go as fast as it went before the debugger was attached.
 
It would be super awesome to be able to do the same in VSCode, and I think – since the code is already there in the ruby-debug-ide gem – it should be manageable to do. Also, this will solve any problem people have with the debugger not finding the right environment (e.g. with rvm in a custom gemset).

rebornix:
VS Code has the process picker UI so the only missing part is this extension :) Nice suggestion.

bjeanes:
Perhaps this might fix #163?

wingrunr21:
Closing for issue cleanup. Apologies if this is still an issue. We are working to improve the core extension experience.

firelizzard18:
@wingrunr21 This is still an issue. Please fix it. It's a really simple change:

```diff
diff --git a/packages/vscode-ruby-debugger/src/main.ts b/packages/vscode-ruby-debugger/src/main.ts
index bd3ec94..e859e13 100644
--- a/packages/vscode-ruby-debugger/src/main.ts
+++ b/packages/vscode-ruby-debugger/src/main.ts
@@ -508,7 +508,11 @@ class RubyDebugSession extends DebugSession {
 
     protected disconnectRequest(response: DebugProtocol.DisconnectResponse, args: DebugProtocol.DisconnectArguments) {
         if (this.rubyProcess.state !== SocketClientState.closed) {
-            this.rubyProcess.Run('quit');
+            if (this.debugMode === Mode.launch) {
+                this.rubyProcess.Run('quit');
+            } else {
+                this.rubyProcess.Run('detach');
+            }
         }
         this.sendResponse(response);
     }
```

pelletencate:
@firelizzard18 This wasn't a bug, this was a feature request. You've commented in the wrong issue.

@wingrunr21 Can we reopen this one please? I really think this is something that's going to be useful for virtually everyone who uses Ruby with resident processes (incl. all Rails devs)

PR #394, opened 2018-10-08
Allow for detaching of debugger

jamesmosier:
Building off of #333, I added a settings option `detachDebugger` which will allow a user to chose to just detach instead of quitting a debugging process. This helps to make sure your Rails server doesn't get shut down whenever you quit the debugger.

If/when this PR is merged, I will update the [wiki page](https://github.com/rubyide/vscode-ruby/wiki/3.-Attaching-to-a-debugger) to reflect the new option. 

- [x] The build passes
- [x] TSLint is mostly happy
- [x] Prettier has been run

### DIFF
--- a/src/debugger/main.ts
+++ b/src/debugger/main.ts
@@ -1,5 +1,6 @@
 'use strict';
 
+import { workspace } from 'vscode';
 import {DebugSession, InitializedEvent, TerminatedEvent, StoppedEvent, BreakpointEvent, OutputEvent, Thread, StackFrame, Scope, Source, Handles, Breakpoint} from 'vscode-debugadapter';
 import {DebugProtocol} from 'vscode-debugprotocol';
 import {readFileSync,existsSync} from 'fs';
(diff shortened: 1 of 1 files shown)


wingrunr21:
Have you verified that you can re-attach to a detached debugger?

jamez14:
Ahh I didn't try that. Unfortunately no, it does not. I believe it is plagued by the same issue mentioned in #125

wingrunr21:
I'm still not sure I like this change. Here's what the VSCode docs say on this:

```
VS Code terminates a debug session with the disconnect request. If the debug target was 'launched' disconnect is expected to terminate the target program (even forcefully if necessary). If the debug target has been 'attached' initially, disconnect should detach it from the target (so that it will continue to run).
```

So this change goes against their debugger recommendations. In addition, what workflow would one use to either reattach to the process (I do need to fix it via #125 but let's assume for now I have not) or kill the process? Wouldn't starting a new debug session spawn a new Ruby process (or another Rails server)?

JesseObrien:
What's the reasoning for this not being worked on still? Because it won't re-attach to the process?

The change seems to be **in line** with their debugger recommendations, if it **attaches** to a debugger, then it should **detach**, not send a `quit` command. Right now it's killing the rdebug-ide process entirely when you disconnect.

wingrunr21:
That's not what this PR does. The debugger (and its behavior) IS being worked on

PR #399, opened 2018-10-23
Call the async-callback to clear parseQueue of items

### LINKED ISSUE #398: Adding new method definition does not add the method to the symbol list
### Your environment

- `vscode-ruby` version: 0.20.0
- Ruby version: 2.3.3
- Ruby version manager (if any): RVM
- VS Code version: 1.28.2
- Operating System: Mac OS X
- Using language server? No

### Expected behavior
When I add a new method in one of my Ruby files and save the file, the new method's name should be available in the symbols list.

### Actual behavior
This works for a while after starting VS Code but then stops working entirely. I.e. new methods don't appear in the symbols list AND the locations of other methods aren't updated, so jumping to the existing symbols is off by a few lines, depending on how much code was added.

I already debugged this issue and the problem is that the `parseQueue` in locate.js never clears items (it doesn't call the async-callback) and therefore becomes saturated and never executes later tasks. The fix is to just call the callback, I'll provide a PR.

peret:
Fixes #398

Not calling the callback caused a bug that meant that after a few tasks
were enqueued on the parseQueue (i.e. a few change-/create-events), all
incoming tasks were queued and never executed. That lead to newly added
symbols not being added to the symbols list and the locations of all
symbols after that being off.

- [x] The build passes
- [x] TSLint is mostly happy
- [ ] Prettier has been run

### DIFF
--- a/src/locate/locate.js
+++ b/src/locate/locate.js
@@ -67,7 +67,7 @@ export class Locate {
 		this.tree = {};
 		this.walkQueue = this._createWalkQueue();
 		this.walkPromise = null;
-		this.parseQueue = async.queue((task, callback) => task().then(() => callback), SINGLE_FILE_PARSE_CONCURRENCY);
+		this.parseQueue = async.queue((task, callback) => task().then(() => callback()), SINGLE_FILE_PARSE_CONCURRENCY);
 	}
 	listInFile(absPath) {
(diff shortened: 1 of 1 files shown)


wingrunr21:
Thanks!

PR #405, opened 2018-11-15
Multi-root support, lint support, DocumentSymbol support, and more!

wingrunr21:
Quick feature overview:

* Adds [multi-root](https://code.visualstudio.com/docs/editor/multi-root-workspaces) support
* Adds diagnostics support (eg lint) with [rubocop](https://github.com/rubocop-hq/rubocop), [reek](https://github.com/troessner/reek), and [standard](https://github.com/testdouble/standard)
* Formatting support (both document and selection) for [rubocop](https://github.com/rubocop-hq/rubocop), [standard](https://github.com/testdouble/standard), and [rufo](https://github.com/ruby-formatter/rufo)
* Adds considerably better folding range support <add before/after here>
* Adds `DocumentSymbol` support which enables the Outline view <add screenshot here>
* Adds out of the box environment detection for RVM, rbenv, chruby, etc environments. If your login shell sets things up correctly, VSCode should set things up correctly


wingrunr21:
Note to watchers: this isn’t dead. I’ve been working through a bug with the shell spawning code that brings down the whole language server. It’s proven to be quite difficult to nail down. I appreciate your patience.

connorshea:
@wingrunr21 is there a reason you did this as one big PR rather than smaller separate PRs?

Thank you for the work on this, I write Ruby in VSCode regularly and this extension is super useful! Best of luck nailing down that bug ❤️

wingrunr21:
@connorshea uhh, not managing my own scope creep! 😊 

More accurately I worked for awhile on a backing structure that would hold up to the various requirements of the extension. That required doing whole or partial implementations for a lot of the pieces to poke holes in things.

wingrunr21:
Update: finally worked through and traced the bug. It ended up being an open bug in Electron where it would throw an uncatchable SIGPIPE when sending stdin to an incorrectly spawned process. It was the same problem that I had looked at (with different symptoms) about six months ago.

What this means is I can now finish up remaining work on this PR and get it released!

wingrunr21:
I will be writing up some docs + a few additional cleanup items and will release this as 0.22.0 tonight/tomorrow

ecbrodie:
@wingrunr21 I would like to inquire about whether this should be considered version 1.0.0.instead of another prerelease version bump. This is a significant rewrite of a majority of this application. Also, assuming vscode-ruby follows semver, anything before 1.0.0 is considered prereleased; this extension has been publicly released for a while now.

wingrunr21:
It will be released as `0.22.0`.

A) VSCode does not have the ability to do pre-release extensions. The only way for an extension to be used is for users do to manual installation or to not really follow semver.
B) There's a good chunk of the extension that doesn't work correctly (like debugging)

I'd like to release v1 when the extension is in a good place. This is a major step in that direction but it isn't there yet.

ecbrodie:
Okay fair enough, I respect that.

connorshea:
👏 👏 👏 👏 👏 👏 

@wingrunr21 nice work!

trkoch:
Should it be possible to have a `.rubcop.yml` in a sub-directory and have VSCode automatically detect it? Are path entries (e.g. ignores) relative to this file? I guess I'm missing something?

wingrunr21:
Everything is run with a `cwd` of the workspace root and/or the file's directory for single files. This would be the same behavior as running RuboCop from the root of your Rails app. Otherwise everything is done via RuboCop itself.

The extension only passes configuration to RuboCop via command line arguments. Otherwise, anything specific via `.rubocop.yml` or other config files are handled by RuboCop.

PR #452, opened 2019-03-13
Added new option to keep ruby process when disconnect request

FabioMR:
My team has been using the debugger with Rails, and every time we disconnect the debugger, the extension quits the ruby process, so my rails server go off as well.
We use the extension in some big projects, and it takes some time to start de server again.

This changes will help us a lot, I probably other users as well.

- [x] The build passes
- [x] TSLint is mostly happy
- [x] Prettier has been run

### DIFF
--- a/package.json
+++ b/package.json
@@ -685,6 +685,11 @@
 								"type": "boolean",
 								"description": "Show output of the debugger in the console.",
 								"default": false
+							},
+							"keepRubyProcessOnDisconnect": {
+								"type": "boolean",
+								"description": "Keep ruby process on disconnect request.",
+								"default": false
 							}
 						}
 					}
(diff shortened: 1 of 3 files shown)


wingrunr21:
Hi,

Thanks for the contribution.

I have the same feedback here as I do for #394: can you reattach to the debugger after detaching?

Also, this config option will be ignored/deprecated when I finish my work on the new debugger. The new debugger will properly honor launch vs attach requests based on `launch.json` . I just want to flag that. I will not be pulling support for this option into the new debugger so it is possible your configs will break down the line if your `launch.json` entries do not align.

FabioMR:
@wingrunr21 

Reattaching is working!

About the new debugger, no problem, when time comes, I can manage it.

FabioMR:
@wingrunr21 Sorry, about the reattaching, I tested again, and it's really a problem!
I will try to fix it.

ducdigital:
Will this be merged to master? I am facing the same issue

wingrunr21:
If the PR gets fixed yes

PR #456, opened 2019-03-16
Fix automatic Rake task detection.

### LINKED ISSUE #273: Can't get rake auto-detection to work
### Your environment

- `vscode-ruby` version: 0.16.0
- Ruby version: 2.4.2
- VS Code version: 1.20.1
- Operating System: macOS
- Hardware (optional): 

I'm using bundler and rbenv.

Make sure you have `ruby`, `ruby-debug-ide` and `ruby-debug-basex19` installed before submitting your issue -- thank you !

### Expected behavior
According to #192 rake tasks should be detected.
When I select Taks => Run Task I expect rake tasks to be auto detected, so I can choose tasks like `rake routes`.

### Actual behavior
No tasks are detected.

### Steps to reproduce the problem
Open a rails project.
Select Taks => Run Task

connorshea:
Previously, the globPattern used in rake.ts was failing to find Rakefiles if they were named `Rakefile, without the `.rb` extension. This fixes the globPattern so now it matches in that case.

This has the downside of also matching any filename that starts with `Rakefile`, e.g. `Rakefile2` or `Rakefileadsadadsad`, but I'm not sure of a great way to fix this without that downside.

If you select `Task: Run Task` in the Command Palette you can see the difference (note the workspace needs to have a file named `Rakefile`, `Rakefile.rb` will be detected without this change).

Before:

Screen Shot 2019-03-15 at 10 19 23 PM

After:
Screen Shot 2019-03-15 at 10 15 51 PM

This fixes #273.

- [x] The build passes
- [x] TSLint is mostly happy
- [x] Prettier has been run

### DIFF
--- a/src/task/rake.ts
+++ b/src/task/rake.ts
@@ -9,12 +9,12 @@ let rakeFiles: Set<vscode.Uri> = new Set<vscode.Uri>();
 
 export async function registerTaskProvider(ctx: vscode.ExtensionContext) {
 	let rakePromise: Thenable<vscode.Task[]> | undefined = undefined;
-	let files = await vscode.workspace.findFiles("**/[rR]akefile{,.rb}");
+	let files = await vscode.workspace.findFiles("**/[rR]akefile{*,.rb}");
 	for (let i = 0; i < files.length; i++) {
 		rakeFiles.add(files[i]);
 	}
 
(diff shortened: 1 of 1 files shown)


connorshea:
@wingrunr21 sorry for the ping, can this fix be merged? :)

connorshea:
Thanks!

PR #480, opened 2019-05-09
Node Runtime

rebornix:
Fix https://github.com/microsoft/vscode-remote-release/issues/153. 

VS Code released [Remote Development](https://code.visualstudio.com/docs/remote/remote-overview) recently, which allows users to connect VS Code to an isolated remote environment, e.g., ssh box, container or WSL. As remote environment is using Nodejs instead of Electron patched Nodejs, current prebuilt tree sitter binaries won't work properly.

To fix this issue 

* This PR added check for host environment and fetch corresponding prebuilt binaries based on both platform/arch and Node runtime
* `tree-sitter-node` already has prebuilt for Node so we want to do the same thing for `tree-sitter-ruby`, here is the PR for it https://github.com/tree-sitter/tree-sitter-ruby/pull/106

### DIFF
--- a/client/src/WorkspaceRubyEnvironment.ts
+++ b/client/src/WorkspaceRubyEnvironment.ts
@@ -3,6 +3,7 @@ import {
 	LanguageClient,
 	RequestType,
 	StaticFeature,
+	InitializeParams,
 } from 'vscode-languageclient';
 import { Uri, workspace, WorkspaceFolder } from 'vscode';
 import { loadEnv, IEnvironment } from './util/env';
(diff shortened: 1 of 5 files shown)


wingrunr21:
I'll look at this tomorrow probably. I'll need to have a new version cut in `tree-sitter-ruby` first though as the prebuilt binaries won't be there until there's a new version (independent of the PR).

wingrunr21:
I took a look at this over the weekend. I'll merge this and cut a release so as to at least prevent the crash, but there's going to be some additional work required to fully support remote environments.

rebornix:
> there's going to be some additional work required to fully support remote environments

@wingrunr21 let me know if you need any information from me for the remote support.

connorshea:
FYI a new release of tree-sitter-ruby was cut

PR #494, opened 2019-06-19
Pass LANG environment variable to language server

### LINKED ISSUE #493: LANG env variable needed for proper Rufo formatting
### Your environment

- `vscode-ruby` version: 0.23.0
- Ruby version: 2.4.6
- Ruby version manager (if any): asdf
- VS Code version: 1.35.1 
- Operating System: Fedora Linux
- Using language server? (eg `useLanguageServer` is true?) yes

### Expected behavior

Formatting Ruby files with rufo as formatter should work if the file contains multibyte utf8 characters like `"“”"`

### Actual behavior

It gives an error:

```
Format: executing rufo --filename=/path/to/file.rb -x...
[Error - 11:12:15] Request textDocument/formatting failed.
  Message: Request textDocument/formatting failed with message: 
    unable to execute rufo --filename=/path/to/file.rb -x:
    Error: Failed with exit code: 1 - unknown error
  Code: -32603 
```

When running Rufo with the same arguments in the command line on the file content, it works:

```
cat /path/to/file.rb | rufo --filename=/path/to/file.rb -x
```

 When running the same command with the `LANG` env variable unset an error occurs:

```
LANG= cat /path/to/file.rb | rufo --filename=/path/to/file.rb -x
# Error: ./path/to/file.rb has syntax errors
```

So the reason seems to be, that the `LANG` environment variable is not set to something with utf8 support like `C.UTF-8`.

pmk1c:
The LANG env variable is needed for rufo, so that it is able to parse ruby files with multibyte characters.
Since it should be available and correct in most Mac and Linux environments we only need to pass the host value.

I tested it and this resolves #493 for me.

- [x] The build passes
- [x] TSLint is mostly happy
- [x] Prettier has been run

### DIFF
--- a/client/src/util/env.ts
+++ b/client/src/util/env.ts
@@ -92,6 +92,7 @@ const RUBY_ENVIRONMENT_VARIABLES = [
 	'GEM_ROOT',
 	'HOME',
 	'RUBOCOP_OPTS',
+	'LANG',
 ];
 
 export interface IEnvironment {


wingrunr21:
This should be fine. When CI passes I'll merge it

PR #497, opened 2019-06-24
Add rake task definition

just806me:
Add missing task definition for rake tasks. This allows some customization of auto-detected tasks (e.g never scan the task output). Without task definition VSCode shows 'Value is not accepted' error and ignores that task.

image

- [x] The build passes
- [x] TSLint is mostly happy
- [ ] Prettier has been run

### DIFF
--- a/package.json
+++ b/package.json
@@ -690,6 +690,24 @@
 					}
 				}
 			}
+		],
+		"taskDefinitions": [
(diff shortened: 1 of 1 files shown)


wingrunr21:
Great! Will merge when CI finishes

Issue #519, opened 2019-08-21
Formatting using rubocop garbles document


tkoenig:
### Your environment

- `vscode-ruby` version: 0.24.2
- Ruby version: 2.6.2
- Ruby version manager (if any):
- VS Code version: 1.37.1
- Operating System: macOS Mojave 10.14.5 
- Using language server? (eg `useLanguageServer` is true?) Yes

### Expected behavior

It should format correctly the following file using rubocop:

Input:

```ruby
# frozen_string_literal: true

puts "this is a test"
puts "\n=========================================================="
puts "okokokok"

```

Expected output (works when using rubocop standalone):

```ruby
# frozen_string_literal: true

puts 'this is a test'
puts "\n=========================================================="
puts 'okokokok'
```
### Actual behavior

The actual output when calling Format document removes most of the document: 

```
"
puts 'okokokok'
```

### More info

It looks like the issue is when parsing the output from `rubocop` it 
thinks that the line with multiple `====` is the marker from rubocop when the actual file content starts.

```
~ ruby-2.6.2 🍔  cat test-rubocop.rb | rubocop -s test-rubocop.rb -a 
Inspecting 1 file
C

Offenses:

test-rubocop.rb:1:1: C: Naming/FileName: The name of this source file (test-rubocop.rb) should use snake_case.
# frozen_string_literal: true
^
test-rubocop.rb:3:6: C: [Corrected] Style/StringLiterals: Prefer single-quoted strings when you don't need string interpolation or special symbols.
puts "this is a test"
     ^^^^^^^^^^^^^^^^
test-rubocop.rb:5:6: C: [Corrected] Style/StringLiterals: Prefer single-quoted strings when you don't need string interpolation or special symbols.
puts "okokokok"
     ^^^^^^^^^^

1 file inspected, 3 offenses detected, 2 offenses corrected
====================
# frozen_string_literal: true

puts 'this is a test'
puts "\n=========================================================="
puts 'okokokok'
```

PR #520, opened 2019-08-21
Fix for detecting rubocop output delimiter

tkoenig:
This would be a quick fix for #519. (I tested this quickly modifying the extensions js file)

Given more thought it might be better to get the **first** instance of `FORMATTED_OUTPUT_DELIMITER`and not the last one to detect the end of rubocop diagnostics?

*Description of change and why it was needed here*

- [] The build passes
- [] TSLint is mostly happy
- [] Prettier has been run


wingrunr21:
Hi,

I resolved this in v0.25.0. This PR was still marked as draft and the `\n` had the potential to not work on Windows. Thanks for the initial bug report!

PR #532, opened 2019-09-14
Add chefignore as a valid Ruby file

tas50:
It's ruby even though there's no extension and it should get all the bells and whistles of this extension.


wingrunr21:
You sure a chefignore file is Ruby? Its syntax looks more like a gitignore file.

wingrunr21:
Closing in favor of 79d6435

Issue #533, opened 2019-09-17
Language server crashes on startup


AlexVPopov:
### Your environment

- `vscode-ruby` version: ﻿0.25.3
- Ruby version: 2.6.0
- Ruby version manager (if any): rbenv
- VS Code version: 1.38.1
- Operating System: macOS 10.14.6
- Using language server? (eg `useLanguageServer` is true?)

```json
{
  "ruby.codeCompletion": "rcodetools",
  "ruby.intellisense": "rubyLocate",
  "ruby.lint": {
    "rubocop": true
  },
  "ruby.useBundler": true,
  "ruby.useLanguageServer": true,
  "ruby.format": "rubocop",
  "[ruby]": {
    "editor.defaultFormatter": "misogi.ruby-rubocop"
  }
}
```

With all extentions disabled and only vscode-ruby enabled, when I open VS Code:

### Expected behavior
I expect the language server to start.

### Actual behavior
The language server crashes 3 times:

```
[Info  - 17:43:16] Initializing Ruby language server...
Loading Ruby tree-sitter syntax from /Users/xxxxxxxxxxxxxxxx/.vscode/extensions/rebornix.ruby-0.25.3/dist/server/tree-sitter-ruby.wasm
/Users/xxxxxxxxxxxxxxxx/.vscode/extensions/rebornix.ruby-0.25.3/dist/server/index.js:16
[...]

PR #539, opened 2019-10-02
Adds rubocop config file param

Maweypeyyu:
I required to overwrite the config file for Rubocop ( via the `-c` cli argument ) and added a config option.

- [x] The build passes
- [x] TSLint is mostly happy
- [x] Prettier has been run
- [x] tested extension (only was able to test without languageServer because it keeps crashing on my ubuntu (same error message as in #533)

### DIFF
--- a/docs/linting.md
+++ b/docs/linting.md
@@ -42,13 +42,14 @@ or
 ```json
 "ruby.lint": {
   "rubocop": {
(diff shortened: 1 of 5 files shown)


Maweypeyyu:
I'll update the PR to resolve the new conflicts if there's interest in merging this in.

PR #550, opened 2019-10-19
Bugfix: Filter out linters that are disabled in the configuration.

### LINKED ISSUE #531: Unable to disable linter
### Your environment

- `vscode-ruby` version: `0.25.3`
- Ruby version: `2.6.2`
- Ruby version manager (if any): `rbenv`
- VS Code version: `1.38.1`
- Operating System: macOS
- Using language server? (eg `useLanguageServer` is true?) Yes

### Expected behavior

With the following in the user settings.json:

```json
"ruby.lint": {
  "rubocop": true
}
```

The following in the workspace settings.json should disable RuboCop:

```json
"ruby.lint": {
  "rubocop": false
}
```

### Actual behavior

RuboCop still runs.

peret:
Fixes #531.

The getLinter() function just passes on the linter config (if it's an object), or continues with an empty object if it's any other values. This leads to unexpected behavior when setting a specific linter to `false` in the config. I.e. the linter is not disabled, as one would expect, but an empty linter configuration is used.

This PR fixes the issue by explicitly filtering linter configs that are `false` before passing them on to getLinter(). 

- [X] The build passes
- [X] TSLint is mostly happy
- [X] Prettier has been run

### DIFF
--- a/packages/language-server-ruby/src/Linter.ts
+++ b/packages/language-server-ruby/src/Linter.ts
@@ -52,7 +52,9 @@ function lint(document: TextDocument): Observable<LintResult> {
 			);
 		}),
 		mergeMap(({ config, env }) => {
-			const linters = Object.keys(config.lint).map(l => getLinter(l, document, env, config).lint());
+			const linters = Object.keys(config.lint)
+				.filter(l => config.lint[l] !== false)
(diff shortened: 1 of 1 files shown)


wingrunr21:
Thanks. This was on my list to look into this weekend but you beat me to it!

Issue #537, opened 2019-09-22
Syntax Highlighting Breaks via Remote-WSL


ecielam:
### Your environment

- `vscode-ruby` version: 0.25.3
- Ruby version: 
- Ruby version manager (if any): No
- VS Code version: Latest
- Operating System: Win 10 + WSL
- Using language server? (eg `useLanguageServer` is true?) Yes

### Expected behavior

When I am running locally, syntax highlighting is correct:

![image](https://user-images.githubusercontent.com/1396086/65390335-a215d380-dd12-11e9-80d7-5c243d7b7dbc.png)

### Actual behavior

If I am running via Remote-WSL, syntax highlighting is broken.

![image](https://user-images.githubusercontent.com/1396086/65390373-d4bfcc00-dd12-11e9-80ce-62348f89ca5f.png)

No errors in the language server console.

wingrunr21:
This has been resolved in `v0.26.0`

Issue #483, opened 2019-05-12
Symbol strings (ex. :abc) via "Remote - SSH" are not correctly highlighted


danglduy:
### Your environment

- `vscode-ruby` version: 0.22.3
- Ruby version: 2.6.3
- Ruby version manager (if any): rbenv
- VS Code version: 1.34.10-insider
- Operating System: Windows
- Using language server? Solargraph
- Using via "Remote - SSH" remote environment _([Link](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-ssh))_

### Expected behavior
Symbol strings syntax is correctly highlighted
### Actual behavior
Symbol strings syntax is incorrectly highlighted

In the picture below `:exception` should be highlighted
![image](https://user-images.githubusercontent.com/38782010/57583136-aa096f80-74f7-11e9-9be2-c14d0353659d.png)

connorshea:
I assume this is the same problem as https://github.com/microsoft/vscode-remote-release/issues/153

danglduy:
I think these are not the same problems. Ruby symbols are correctly highlighted without any language servers anyway.

minkir014:
Have you tried that on an ordinary vs code?

danglduy:
@minkir014 "Remote - SSH" extension does only work on VS Code Insider I'm afraid.

wingrunr21:
hi @danglduy 

Can you use the `Developer: Inspect TM Scopes` tool to tell me what scopes are being matched for the `with: :exception` part of that code?

The `with:` should be matching the grammar with a scope of `constant.language.symbol.hashkey.ruby` and `:exception` should be `constant.language.symbole.ruby`.

danglduy:
@wingrunr21 I inspected `with:` and `:exception`
![image](https://user-images.githubusercontent.com/38782010/57635969-764f4800-75d2-11e9-9103-74aeb280004b.png)
![image](https://user-images.githubusercontent.com/38782010/57636006-86ffbe00-75d2-11e9-9ddb-861d52816fd6.png)

wingrunr21:
Ok. What theme are you using? The Ruby grammar is correctly matching the TM scopes but it appears your theme does not have support.

danglduy:
@wingrunr21 I am using the default theme.

wingrunr21:
Which is what? `Dark+`?

danglduy:
Yes it’s Dark+

wingrunr21:
[...]

PR #551, opened 2019-10-19
Remote environment support

### LINKED ISSUE #488: Move environment detection to server
I made the decision to run environment detection on the client originally as that seemed to be the direction remote workspaces were headed. Well, I guessed wrong on that. Environment detection needs to be moved to the server so that remote development environments work correctly.

### LINKED ISSUE #490: Flag machine specific configurations like paths as machine scoped
I am developer from VS Code and we introduced Machine Scope for configurations in [1.34](https://code.visualstudio.com/updates/v1_34#_machinespecific-settings) which will help in [remote set ups](https://code.visualstudio.com/docs/remote/remote-overview).

I would like to encourage extension authors to adopt to machine scope if you have any settings that you think they are scoped to machine like `ruby.interpreter.commandPath` 

For example, If you have settings that allow users to customize an executable path that exists on the machine in which extension is running, then you can tag such settings as `machine` scoped. This will let VS Code to read this setting from the location where the extension is running.

- When extension is running locally, VS Code reads this setting **only** from User settings on the same machine.
- When extension is running remotely, VS Code reads this setting **only** from Remote settings on the remote machine.

If the extension is running remotely in a [remote set up](https://code.visualstudio.com/docs/remote/remote-overview), then this will help the extension not to read paths configured in local machine user settings and bail as the paths not exist in the remote machine.

Note that machine scoped settings are overridable only in user or remote settings.

### LINKED ISSUE #523: Support fish shell in environment detection
### Your environment

- `vscode-ruby` version: 0.24.2
- Ruby version: 2.6.3
- Ruby version manager (if any): chruby
- VS Code version: 1.37.1
- Operating System: macOS 10.14.6
- Using language server? Yes
- Login shell: /usr/local/bin/fish

### Expected behavior

Enabling ruby linter via rubocop (using bundler) should check file without errors.

### Actual behavior

Language server throws following error:
```
[Info  - 2:33:48 PM] Initializing Ruby language server...
[Info  - 2:33:48 PM] Rebuilding tree-sitter for local Electron version
[Info  - 2:33:48 PM] Rebuild succeeded!
Lint: executing bundle exec rubocop -s /Users/e1senh0rn/Projects/test/app.rb -f json...
/Users/e1senh0rn/.rubies/ruby-2.6.3/lib/ruby/2.6.0/rubygems.rb:283:in `find_spec_for_exe': Could not find 'bundler' (2.0.1) required by your /Users/e1senh0rn/Projects/test/Gemfile.lock. (Gem::GemNotFoundException)
To update to the latest version installed on your system, run `bundle update --bundler`.
To install the missing version, run `gem install bundler:2.0.1`

	from /Users/e1senh0rn/.rubies/ruby-2.6.3/lib/ruby/2.6.0/rubygems.rb:302:in `activate_bin_path'
	from /Users/e1senh0rn/.gem/ruby/2.6.3/bin/bundle:23:in `<main>'

Lint: Received invalid JSON from rubocop:
```

Running same command from terminal works just fine.

I investigated an error, and it really has nothing to do with bundler itself.

* Ruby version is 2.6.3. ABI version is 2.6.0. Rubygems by default tries to lookup gems in `~/.gem/ruby/<ABI-version>`. I have non-bundled gems installed in ~/.gem/ruby/2.6.3, which prevents it from finding bundler.
* This can be adjusted with `GEM_PATH` environment variable.
* chruby actually sets variables correctly within shell, hence no errors while running it from terminal.
* vscode-ruby launches login shell in workspace directory to figure out environment variables.
* To launch shell it creates shell shim to list variables.
* I'm using fish shell, and resulting shim looks like this ( ~/.vscode/extensions/rebornix.ruby-0.24.2/client/out/util/shims/env.fish.sh):
```
#!/usr/local/bin/fish -i

export
```
* In fish `export` lists variables as `VARIABLE value`.
* `processExportLine` expects variables and values to be separated by `=`.
* This produces completely blank environment (`{}`) when launching ruby command.

This behavior will cause error with non-bash shells (`fish`, `elvish`, `csh`, `tcsh`, ...).

Affected code: https://github.com/rubyide/vscode-ruby/blob/master/packages/vscode-ruby/src/util/env.ts#L17

Could it be better to invoke `/usr/bin/env` instead of `export`?

wingrunr21:
* Flags config options as `machine` or `machine-overridable`. Resolves #490
* Officially supports remote environments. Resolves #551, #483, #537 
* Migrates environment detection into the language server. Resolves #488 
* Uses `vscode-ruby-common` for environment detection. Resolves #523


connorshea:
❤️ thanks!

PR #553, opened 2019-10-25
Add block snippets

wdspkr:
This adds snippets for do..end blocks, do..end blocks with variables and
blocks with curly braces. Those snippets are super handy and IMHO essential. They are derived from the [atom ruby project](https://github.com/atom/language-ruby) and can be found here: https://github.com/atom/language-ruby/blob/master/snippets/language-ruby.cson#L153

- [x] The build passes
- [x] TSLint is mostly happy
- [x] Prettier has been run

### DIFF
--- a/snippets/ruby.json
+++ b/snippets/ruby.json
@@ -185,5 +185,28 @@
 			"\tend",
 			"end"
 		]
+	},
+	"Insert do … end block": {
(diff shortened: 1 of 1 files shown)


wingrunr21:
Thanks! I cherry-picked your commit into master

PR #575, opened 2019-12-24
Add Stdin-filename arg to Reek linter

guzart:
Passes an `--stdin-filename` argument to Reek when linting files.

This gives Reek context about the piped contents and supports using the [`directories` configuration](https://github.com/troessner/reek#working-with-rails). 

Caveat:
Because this change sends the absolute path to the file, the configuration in directories must be prefixed with `**` for it to work. e.g.
```yml
directories:
  '**/app/controllers':
    IrresponsibleModule:
      enabled: false
```

### DIFF
--- a/packages/language-server-ruby/src/linters/Reek.ts
+++ b/packages/language-server-ruby/src/linters/Reek.ts
@@ -1,3 +1,4 @@
+import { URI } from 'vscode-uri';
 import BaseLinter from './BaseLinter';
 import { Diagnostic, DiagnosticSeverity } from 'vscode-languageserver';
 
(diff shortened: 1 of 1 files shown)


minkir014:
When this PR will be merged???

leonelgalan:
> Caveat:
> Because this change sends the absolute path to the file, the configuration in directories must be prefixed with ** for it to work. e.g.

Because of this change and before I add the ** the linter from the extension will show smells that won't show when I run:

```sh
bundle exec reek folder/file.rb
```

If I prefix directories with ** as suggested, the smells don't show on vscode-ruby, but show up when I run reek on it's own.

Issue #627, opened 2020-06-03
Support latest versions of rubyfmt


wadetandy:
### Your environment

- `vscode-ruby` version: 0.27.0
- Ruby version: any
- Ruby version manager (if any): rvm
- VS Code version: 1.45.1
- Operating System: OSX
- Using language server? yes

### Expected behavior

I have installed the latest version of rubyfmt per the [readme instructions](https://github.com/penelopezone/rubyfmt#how-do-i-use-it), so it is available on my bash `PATH` as `rubyfmt`. I would expect this command to get run when the vscode formatter runs.

### Actual behavior

Per [this file](https://github.com/rubyide/vscode-ruby/blob/master/packages/language-server-ruby/src/formatters/RubyFMT.ts#L4), vscode-ruby embeds a much older version of rubyfmt, which the author notes as broken in https://github.com/penelopezone/rubyfmt/pull/201.

wingrunr21:
I haven't looked at `rubyfmt` in a while. It looks like it is getting transitioned to something I shouldn't vendor any longer. I'll have to do some work to redo the integration to allow it to be external similar to other formatters.

wadetandy:
@wingrunr21 Opened #628 to address this change.

wingrunr21:
`v0.28.0` contains this support

PR #628, opened 2020-06-07
Get modern rubyfmt working

wadetandy:
Per #627, the previous behavior of the rubyfmt formatter class included
an embedded version of the rubyfmt script. This isn't really a workable
solution anymore as the program has evolved beyond a simple ruby script
to include native modules.  This instead changes the formatter class to
rely on a `rubyfmt` executable on the user's `$PATH`.  Because rubyfmt
is not currently installable via bundler and is in a pre-release state,
users are expected to have followed the necessary installation
instructions from the [project
README](https://github.com/penelopezone/rubyfmt/blob/master/README.md#how-do-i-use-it).
The current instructions recommend adding an alias, but vscode will not
respect that (or at least vscode-ruby doesn't). Instead I have a
[commit](https://github.com/penelopezone/rubyfmt/pull/201/commits/19d5946a43b77de4be1083b3097b632dd79ab1bb)
submitted in [a PR](https://github.com/penelopezone/rubyfmt/pull/201)
which will make it easier to install an actual executable shim script.

- [x] The build passes
- [ ] TSLint is mostly happy
- [x] Prettier has been run

`yarn lint` is failing on a few things, but nothing in the changed files from this PR.


wingrunr21:
Hey, thanks for this. Looks good overall. I do prefer using a shim on the Rubyfmt side but if Penelope prefers to direct people to use an alias we could keep the direct `ruby` call. It looks like the invocation is pretty close to the old version just with the addition of the Rust native extension.

The inability to use an alias is linked to node's `spawn`. A bash alias requires an interactive shell in order to be expanded and `spawn` does not support using an interactive shell.

wingrunr21:
@wadetandy if this is ready I'm happy to merge it

wadetandy:
I think this is ready to go with the caveat that it won't really work until the corresponding shim script PR in rubyfmt gets merged, which should hopefully be this week sometime.

wadetandy:
Rubyfmt changes & vscode-ruby usage instructions were merged, so this should be GTG!

PR #640, opened 2020-08-01
Add html basic tags to .erb files

Arthur092:
Hi guys, developing in Ruby I found this helpful extension, but when trying to work with .erb files, didn't find any shortcut or snippet to write html tags easily, so I added the most used html tag in this PR.


minkir014:
When will this be merged?

Issue #658, opened 2020-09-21
@do causes wrong syntax highlighting


Nowaker:
### Your environment

- `vscode-ruby` version: 0.27.0
- Ruby version: 2.6.3p62
- Ruby version manager (if any): rvm
- VS Code version: 1.49.1
- Operating System: Arch Linux
- Using language server? yes

### Actual behavior

![image](https://user-images.githubusercontent.com/611746/93824552-403fc080-fc29-11ea-9323-3daedd24d46d.png)

### Expected behavior

Proper syntax highlighting. `@do` is mistaken for `do`. `@do` is legal instance variable name, not a keyword, so it shouldn't cause this problem.

wingrunr21:
Resolved by #660

PR #660, opened 2020-10-11
Check for : @ and $ before 'do'

graceful-potato:
This is a fix for  #658 
First I just wanted to check that there is a start of line or space char before `do` keyword, but there is some weird cases where you can put `do` right after the method name without space:

```ruby
def +
  yield "test"
end

self.+do |e|
  e.upcase
end
```
I am pretty sure this will work for any of these methods: `+`, `-`, `*`, `/`, `^`, `%`, `&`, `|`
Blocks are rarely used with them and blocks without space between method name and `do` I think almost never.
So I am not sure what is better.

### DIFF
--- a/packages/vscode-ruby/syntaxes/ruby.cson.json
+++ b/packages/vscode-ruby/syntaxes/ruby.cson.json
@@ -2106,7 +2106,7 @@
 			]
 		},
 		{
-			"begin": "(?<={|{\\s|[^A-Za-z0-9_]do|^do|[^A-Za-z0-9_]do\\s|^do\\s)(\\|)",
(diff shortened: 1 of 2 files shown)


wingrunr21:
I'd just make it whatever is going to be right most of the time.

Part of my radio silence on this is converting things over to the `tree-sitter` query support. That should open up the semantic highlighting API that VSCode recently shipped which will hopefully cover the rest of these edge cases.

graceful-potato:
I heard about tree sitter, but I dont know much about it. This is just a quick fix in standard grammar file.

wingrunr21 (approved):

Nowaker:
Thank you!

Issue #451, opened 2019-03-08
Rubocop causes language server to crash on start


peret:
### Your environment

- `vscode-ruby` version: 0.22.3
- Ruby version: 2.5.3
- Ruby version manager (if any): rvm
- VS Code version: 1.32.1
- Operating System: Mac OS X
- Using language server? Yes

### Expected behavior
Language server should start and start linting files with rubocop

### Actual behavior
Sorry to open another one of these, but recently the language server stopped working for me on one of my systems. 

```
Debugger listening on ws://127.0.0.1:6009/150f4a70-2974-4c7e-8591-db1d1be62a0e
For help, see: https://nodejs.org/en/docs/inspector
[Info  - 8:07:25 PM] Initializing Ruby language server...
[Info  - 8:07:25 PM] Rebuilding tree-sitter for local Electron version
[Info  - 8:07:25 PM] Rebuild succeeded!
Changed file
Lint: executing rubocop -s /Users/peter/XXXXXX/test/models/xxxxxx.rb -f json -R...
Lint: Received invalid JSON from rubocop:

Ignoring jaro_winkler-1.5.2 because its extensions are not built.  Try: gem pristine jaro_winkler --version 1.5.2
Ignoring jaro_winkler-1.5.1 because its extensions are not built.  Try: gem pristine jaro_winkler --version 1.5.1
<more lines with more gems here>

/System/Library/Frameworks/Ruby.framework/Versions/2.3/usr/lib/ruby/2.3.0/rubygems/core_ext/kernel_require.rb:120:in `require': cannot load such file -- jaro_winkler (LoadError)
	from /System/Library/Frameworks/Ruby.framework/Versions/2.3/usr/lib/ruby/2.3.0/rubygems/core_ext/kernel_require.rb:120:in `require'
	from /Users/peter/.rvm/gems/ruby-2.5.3@ptcshell/gems/rubocop-0.65.0/lib/rubocop/string_util.rb:3:in `<top (required)>'
	from /Users/peter/.rvm/gems/ruby-2.5.3@ptcshell/gems/rubocop-0.65.0/lib/rubocop.rb:21:in `require_relative'
	from /Users/peter/.rvm/gems/ruby-2.5.3@ptcshell/gems/rubocop-0.65.0/lib/rubocop.rb:21:in `<top (required)>'
[...]

PR #683, opened 2020-11-23
fix: default to login shell (`-l`) when opening a shell

### LINKED ISSUE #620: Ruby version not correctly parsed when using RVM
### Your environment

- `vscode-ruby` version: 0.27.0
- Ruby version: 2.7.1
- Ruby version manager (if any): RVM
- VS Code version: 1.45.1
- Operating System: MacOS 10.15.4
- Using language server? (eg `useLanguageServer` is true in your configuration?) `true`

### Expected behavior

When linting or formatting (like with Rubocop) and using bundler, no errors are outputted and formatting behavior works.

### Actual behavior

**Workspace settings:**
```json
{
  "ruby.useBundler": true,
  "ruby.useLanguageServer": true,
  "ruby.codeCompletion": false,
  "ruby.intellisense": false,
  "ruby.format": "rubocop",
  "ruby.lint": {
    "rubocop": {
      "useBundler": true
    }
  }
}
```

**Directory structure:**
```
|- app.rb
|- Gemfile
|- Gemfile.lock
```

The following error is logged due to the wrong version of Ruby being loaded:
```
Lint: executing bundle exec rubocop -s /Users/admin/my-app/app.rb -f json...
/Users/admin/.rvm/gems/ruby-2.6.3/gems/bundler-2.1.4/lib/bundler/definition.rb:495:in `validate_ruby!': Your Ruby version is 2.6.3, but your Gemfile specified 2.7.1 (Bundler::RubyVersionMismatch)
	from /Users/admin/.rvm/gems/ruby-2.6.3/gems/bundler-2.1.4/lib/bundler/definition.rb:470:in `validate_runtime!'
	from /Users/admin/.rvm/gems/ruby-2.6.3/gems/bundler-2.1.4/lib/bundler.rb:143:in `setup'
	from /Users/admin/.rvm/gems/ruby-2.6.3/gems/bundler-2.1.4/lib/bundler/setup.rb:20:in `block in <top (required)>'
	from /Users/admin/.rvm/gems/ruby-2.6.3/gems/bundler-2.1.4/lib/bundler/ui/shell.rb:136:in `with_level'
	from /Users/admin/.rvm/gems/ruby-2.6.3/gems/bundler-2.1.4/lib/bundler/ui/shell.rb:88:in `silence'
	from /Users/admin/.rvm/gems/ruby-2.6.3/gems/bundler-2.1.4/lib/bundler/setup.rb:20:in `<top (required)>'
	from /Users/admin/.rvm/rubies/ruby-2.6.3/lib/ruby/site_ruby/2.6.0/rubygems/core_ext/kernel_require.rb:54:in `require'
	from /Users/admin/.rvm/rubies/ruby-2.6.3/lib/ruby/site_ruby/2.6.0/rubygems/core_ext/kernel_require.rb:54:in `require'
```

The content of `app.rb` itself is empty, meaning it's not affected by the source code.

### Hypothesis
When RVM is loaded, it aliases the `rvm` command as a function and sets up appropriate environment variables (like `RUBY_VERSION`) which is usually set at a user's `~/.profile` or `~/.bash_profile` dotfile. However those dotfiles are only sourced when running a shell with the login flag `/bin/bash -l`, otherwise RVM isn't loaded at all.

It seems that the fix would be to change this line to also pass the `-l` flag https://github.com/rubyide/vscode-ruby/blob/master/packages/vscode-ruby-common/src/environment.ts#L44.

I haven't tested it yet but it may also require a `cd .` statement to trigger RVM's automatic Ruby loading when changing directories.

nicolasrouanne:
Add the `-l` flag when opening a shell so that dotfiles are loaded. Fixes rvm not being loaded properly.
Fixes #620, #451 

- [] The build passes
- [] TSLint is mostly happy
- [] Prettier has been run

### DIFF
--- a/packages/vscode-ruby-common/src/environment.ts
+++ b/packages/vscode-ruby-common/src/environment.ts
@@ -41,7 +41,7 @@ for name in (set -nx)
 	end
 end`;
 	} else {
-		template = `#!${shell} -i\nexport`;
+		template = `#!${shell} -i -l\nexport`;
 	}
 
 	return template;


mattstrayer:
Can we open this back up, this is a much needed fix to the extension

nicolasrouanne:
Is there any way we could reopen this and merge or get feedback?
IMO it seems like a legit solution, but 🤷🏻‍♂️

BTW test failings seems to be random would be happy to try to make another test run to check if it passes

wingrunr21:
This solution might be "legit" but it also has the potential to make every commend that is executed by the language server considerably slower. An interactive shell is required at a minimum due to most version managers exposing their functionality via shell functions. I'd prefer this is exposed as a configuration option vs requiring all users to load their full shell environment on every invocation.

mattstrayer:
Agree with @wingrunr21 that this fits better behind a config option!

nicolasrouanne:
Agreed for the config option @wingrunr21 . Would you make a PR or guide me a little how to handle it?

wingrunr21:
If you are able to open a PR to do it that'll likely be faster than me getting to it.

All of the work should be doable within the `vscode-ruby-client` package. The language server itself doesn't need to know about this config option since environment detection occurs on the client side.

1. Add the config option to the client `package.json`. Probably makes sense to put that under the `languageServer` key. There are a few other environment detection feature requests to bonus points for the `languageServer.environmentDetection` key being used as the parent for this setting.
2. The [`getShim`](https://github.com/rubyide/vscode-ruby/blob/master/packages/vscode-ruby-client/src/util/env.ts#L43) function needs to be made aware of the config option.

Since the shims themselves are an implementation detail, my recommendation would be to utilize a naming convention on the shim (similar to how the shell name is handled). The shims are only generated if they don't exist so if a naming convention isn't used some other mechanism needs to be used to see if the `-l` option is present in the shim.

TSMMark:
Did a config option get added for this?

gregawoods:
Has an alternative solution to this been added or considered? It feels like a real shame to ignore this, RVM is a very common use case.

Issue #665, opened 2020-10-12
Formatters adding extra newlines on windows if VC Code eol setting is LF insteads of CRLF


nipinium:
### My environment

- `vscode-ruby` version: v0.27.0
- Ruby version: 2.7.2p137
- Ruby version manager (if any): none
- VS Code version: 1.50.0
- Operating System: Windows 10
- Using language server? Yes

### Expected behavior
The formatters work correctly.

### Actual behavior
The formatters keep inserting extra newlines if the default end of line character is LF (`\n`) insteads of CRLF (`\r\n`). I tried ruby, rubyfmt and standardrb, all behaved the same.

Before formatting:
![before](https://user-images.githubusercontent.com/34429224/95766946-37995200-0cde-11eb-80d3-935dd1254f9e.png)

After formatting:
![after](https://user-images.githubusercontent.com/34429224/95766970-408a2380-0cde-11eb-8fae-4cc4ea7066d2.png)

cscorley:
I also observe this issue, but with `"files.eol": "auto"`.  However, with auto set, it is using CRLF for .rb files (indicated in status bar) and changing them (via status bar) to LF also causes this behavior.

Some other observations:

1. Running rubocop from Terminal (pwsh) does not do this: `rubocop -a`
2. Subsequent saves with auto (CRLF), which re-run rubocop from the plugin, will actually revert the line changes on every other save for some files.  So, every save either introduces more new lines or removes them.

cscorley:
This looks related to #644.  

The solution suggested there is to use cmd instead of powershell.  However, switching to `"terminal.integrated.shell.windows": "C:\\Windows\\System32\\cmd.exe"` and `"terminal.integrated.automationShell.windows": "C:\\Windows\\System32\\cmd.exe",` does not fix this.  Using defaults for those settings also does not work, since the default for VS Code is PowerShell now.

nipinium:
I'm using `cmd.exe` by default. Confirmed that it does not resolve the issue.

cscorley:
Me too.  I've tried it several different ways, and observe the same behavior no matter what terminal type I am using and even when using different formatters.  I am thinking it might be either how the edits are actually applied or the diffing engine is not compatible with CLRF line endings.

For the first thought, I think that this form of transformation in the BaseFormatter may be outdated:

[...]

Issue #644, opened 2020-08-06
Format with RuboCop adds line breaks between every line of code


philomory:
### Your environment

- `vscode-ruby` version: 0.27.0
- Ruby version: 2.7.1p83
- Ruby version manager (if any): uru
- VS Code version: 1.47.2
- Operating System: Windows 10
- Using language server? (eg `useLanguageServer` is true in your configuration?): Yes

### Expected behavior

When I run the `Format Document` command with an already-valid document, no changes should be made to the document.

### Actual behavior

Before running `Format Document`:
![image](https://user-images.githubusercontent.com/200123/89485208-f9ab0b00-d73b-11ea-9d01-a884c5c8d60e.png)

After running `Format Document`:
![image](https://user-images.githubusercontent.com/200123/89485243-11828f00-d73c-11ea-99ac-6b7c368497d4.png)

I thought this might be because I'm on Windows but using `LF` rather than `CRLF` for line endings (in both VS Code and `.rubocop.yml`), but, switching everything to `CRLF` didn't actually help.

This feels similar to some weird stuff I've seen happen when using pipes in Powershell, but I haven't dug around into the extension code to see if we're actually doing that on Windows or not.

wingrunr21:
Interesting. Using cmd vs powershell depends on what your default shell is set to. Have you turned on verbose logging and looked at the server logs?

philomory:
@wingrunr21 Sorry if I'm being thick, but, how do you turn on verbose logging? I can't find anything about that in either the extension documentation or in VSCode's documentation. Certainly I've checked the Output pane, but all it says is
```
Format: executing rubocop.bat -s c:\path\to\dummy.rb -a...
undefined
```

reyou:
I think this needs to be investigating. A major issue on Windows / VS Code.

FYI I run following command as installation:

`gem install rubocop`

PR #685, opened 2020-11-27
fix: maintain original EOL after formatting (issue #665)

cscorley:
This implements a workaround for formatters that change the line endings of files.   This change looks to address two problems:

1. When a file in LF mode in VS Code and the formatter outputs CRLF, VS Code will convert the CRs inserted by the TextEdit objects as new LFs, thus producing double lines as seen in #665 and #644 
2. On Windows, Rubocop and Standardrb appear to have bugs that, on CRLF inputs, will produce CRCRLF inputs.  VS Code will convert the CRs inserted by the TextEdit objects as new CRLFs, similarly producing double lines.

Since changing the line ending type of the entire file doesn't seem possible by TextEdits alone, we can instead "undo" the line ending changes the formatters do, while still producing TextEdits of the other changes.

This approach works by determining which line ending the original file was in, and converting the output formatted text to that line ending before diffing the text and generating TextEdits.  

It is entirely possible there is a better approach, but I do not know enough about VS Code extension writing to do better yet.

- [x] The build passes
- [x] TSLint is mostly happy
- [x] Prettier has been run

### DIFF
--- a/packages/language-server-ruby/src/util/spawn.ts
+++ b/packages/language-server-ruby/src/util/spawn.ts
@@ -6,7 +6,7 @@ import log from 'loglevel';
 
 export type SpawnOpts = {
 	stdin?: Observable<string>;
-	encoding?: string;
+	encoding?: BufferEncoding;
 } & SpawnOptions;
 
 export interface SpawnReturn {
--- a/packages/language-server-ruby/package.json
+++ b/packages/language-server-ruby/package.json
@@ -9,7 +9,9 @@
 	"main": "dist/index.js",
 	"dependencies": {
 		"cross-spawn": "^6.0.5",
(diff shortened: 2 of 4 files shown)


packages/language-server-ruby/src/util/spawn.ts
```diff
@@ -6,7 +6,7 @@ import log from 'loglevel';
 
 export type SpawnOpts = {
 	stdin?: Observable<string>;
-	encoding?: string;
+	encoding?: BufferEncoding;
```
cscorley, line 9: I was unable to compile without this change.  Buffer.toString() on L53 was not able to ensure the type could be converted since `string` here could be anything.

floh96:
@wingrunr21 please reopen

Issue #510, opened 2019-07-17
Launching with the debugger does not work if there are disabled breakpoints


AlexMunro:
### Your environment

- `vscode-ruby` version: 0.24.1
- Ruby version: 2.6.3p62
- Ruby version manager (if any): asdf
- VS Code version: 1.36.1
- Operating System: MacOS Mojave 10.14.5
- Using language server? (eg `useLanguageServer` is true?) No

Versions of possibly relevant gems (though updating didn't help in this case):
ruby-debug-ide version (0.6.0)
debase (0.2.3)
debase-ruby_core_source (0.10.5)

### Expected behavior

After adding a breakpoint and disabling it, starting the debugger should work as normal, ignoring the disabled breakpoints.

### Actual behavior

Leaving any disabled breakpoints in prevents the debugger from (properly?) starting. Nothing is written to the debug console, and in the case of the Rails server it does not respond to requests. However, the draggable debug toolbar does appear which responds to stop/restart and trying to launch twice gives the usual error message `There is already a debug configuration <config name> running.`

Enabling or removing the breakpoints and restarting allows the launch configurations to work as normal.

I've observed this behaviour with a Rails project (using the default configuration for launching rails servers) and a plain Ruby project (using the configuration listed here: https://github.com/rubyide/vscode-ruby/wiki/2.-Launching-from-VS-Code).

Obviously there's an easy workaround here, but it took a while to figure out this was the issue so it would be useful to have some kind of warning if a fix isn't feasible.

griest024:
I have the same issue with a rack server. With logging enabled, the only message printed is `Fast Debugger (ruby-debug-ide 0.7.0, debase 0.2.2, file filtering is supported) listens on 127.0.0.1:1234` and puma never launches.

castwide:
Same problem here. I just wanted to mention that launching with disabled breakpoints used to work, but I don't know exactly when it stopped.

iggycoder:
Yep, same problem here.

whitehorsesoft:
Problem still exists, extension 0.25.3, OSS code v 1.39.2, electron 4.2.12

AlexMunro:
Still seeing this issue with debase 0.2.4.1, ruby-debug-ide 0.7.2, Ruby 2.7.1 and vscode-ruby 0.27.0.

whitehorsesoft:
Planning on verifying once have time, to make sure this issue isn't closed inappropriately...

lalunamel:
[...]

Issue #657, opened 2020-09-16
VS code does not send 'start' command when a debugging session begins


lalunamel:
Issue Type: <b>Bug</b>

### Your environment

- `vscode-ruby` version: 0.27.0
- Ruby version: 2.6.6
- Ruby version manager (if any): chruby
- VS Code version: Code 1.49.0 (e790b931385d72cf5669fcefc51cdf65990efa5d, 2020-09-10T17:39:53.251Z)
- Operating System: Mac OS Darwin x64 19.6.0
- Using language server? yes 

Hello! I've been really digging into how VS code does native debugging and I believe I've found a bug!

Unfortunately, I'm not able to reproduce this issue myself - it only appears on my teammate's machine. I've been the one investigating the issues with debugging, so I'm submitting the bug report here. We verified together that we are using the same versions of `vscode-ruby`, ruby, chruby, mac, and vscode.

We've found that VS code does not send the `start` command when a debugging session begins. In our case, that causes execution of the process we're attempting to debug to hang.

## Reproduction steps

Here are the reproduction steps (some of which are probably extraneous to the issue, but they're what I've been doing to repro the issue):

**1. Get `ruby-debug-ide` set up**
We're running ruby 2.6.6, but I doubt that makes a difference
`gem install ruby-debug-ide`
`gem install debase`
These gems provide the ability to debug a ruby executable by starting a server that can communicate with VS code's debugging functionality.

**2. Install the `Ruby` extension in VS code**
You can see below that I'm using version `0.27.0`

**3. Add a new launch config to `launch.json`**
```
{
      "name": "Attach to debug ruby",
      "type": "Ruby",
      "request": "attach",
      "remoteHost": "127.0.0.1",
      "remotePort": "1235",
      "remoteWorkspaceRoot": "${workspaceRoot}",
      "showDebuggerOutput": true
}
```
**4. Start the ruby debug server**
`rdebug-ide --host 0.0.0.0 --port 1235 --debug -- echo "hello"`
`rdebug-ide` is the binary installed with the `ruby-debug-ide` gem
`--host` and `--port` specify the host and port used to connect to the server. Note we're using the same values in the launch config and the command above.
`--debug` tells `ruby-debug-ide` to output verbose logs
` -- ` is a delimiter that specifies that what follows is an executable to run and not another option.
[...]

PR #688, opened 2020-12-23
fix: when a setBreakpoint request comes in, always respond only once

jvilk-stripe:
*Description of change and why it was needed here*

Fixes https://github.com/rubyide/vscode-ruby/issues/510 and https://github.com/rubyide/vscode-ruby/issues/657

This fixes a notorious bug that engineers at Stripe encounter daily: The debugger fails to attach because they have a breakpoint that is unset in a file.

Specifically, the situation occurs when VS Code has one file open that has _all_ unset breakpoints (e.g., there are no active breakpoints in the file). It turns out that VS Code sends out a `setBreakpoints` request with _no_ breakpoints for this file. In this case, the Ruby debug adapter's code would not send a response to this request, so VS Code would wait forever for a response that would never come.

I've updated the code to ensure that it always responds once to this sort of request, and fixes another bug where I think it might respond twice to a single setBreakpoints request.

- [] The build passes
- [] TSLint is mostly happy
- [] Prettier has been run: I did not run prettier because the code I edited has never been prettiered, and I did not want to dirty the diff.

### DIFF
--- a/packages/language-server-ruby/src/util/spawn.ts
+++ b/packages/language-server-ruby/src/util/spawn.ts
@@ -50,7 +50,7 @@ export function spawn<T = string>(
 				if (typeof b === 'string') {
 					chunk = b.toString();
 				} else {
-					chunk = b.toString(optsWithoutStdIn.encoding || 'utf8');
+					chunk = b.toString(<BufferEncoding> optsWithoutStdIn.encoding || 'utf8');
 				}
 			} catch (e) {
 				chunk = `<< Lost chunk of process output for ${cmd} - length was ${b.length}>>`;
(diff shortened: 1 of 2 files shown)


packages/language-server-ruby/src/util/spawn.ts
```diff
@@ -50,7 +50,7 @@ export function spawn<T = string>(
 				if (typeof b === 'string') {
 					chunk = b.toString();
 				} else {
-					chunk = b.toString(optsWithoutStdIn.encoding || 'utf8');
+					chunk = b.toString(<BufferEncoding> optsWithoutStdIn.encoding || 'utf8');
```
jvilk-stripe, line 53: For some reason this code wouldn't typecheck / build w/o this cast.

jvilk-stripe:
@wingrunr21 Hello again friend! Fixing this bug would be really impactful for us at Stripe. Please let me know if there's anything I can do to make the review process easier for you.

wingrunr21 (approved):

jvilk-stripe:
@wingrunr21 exciting! 🎉  do you have plans for a release w/ this fix anytime soon? looks like the last vscode-ruby release was in March or so.

wingrunr21:
Yes, I'm doing a few dependency updates and will get it cut. Today/tomorrow.

PR #749, opened 2021-07-12
feat: add skipFiles and showDebuggerCommands

nagl-stripe:
At Stripe, we use sorbet and other internal utilities that 'pollute' the stack - they create
a large number of frames that are not interesting to our developers when debugging. This commit
adds two features that help avoid those frames.

skipFiles was inspired by the corresponding js debugger feature. It's useful for avoiding
uninteresting frames that are between interesting frames (sorbet falls in to this category).
Skipped files do not appear in stack traces, an the debugger automatically steps through them.

finishFiles is useful for avoiding frames that define the top of a never-interesting call stack.
The debugger automatically steps out of them instead of through them.

showDebuggerCommands is like showDebuggerOutput, but in the other direction. I found it useful
when debugging both features.

- [?] The build passes - the same 3 tests fail with or without this commit
- [?] TSLint is mostly happy
- [ ] Prettier has been run - no, but I did touch things up when I was already changing a line. Running prettier deserves a commit of its own.

### DIFF
--- a/docs/debugger.md
+++ b/docs/debugger.md
@@ -26,6 +26,32 @@ Read following instructions about how to debug ruby/rails/etc locally or remotel
 
 ## Debugger F.A.Q.
 
+Q: What's the difference between 'skipFiles' and 'finishFiles'?  
+A: The debugger will automatically step through skipFiles, whereas it will automatically step out of finishFiles.
+
+Q: What are skipFiles and finishFiles useful for?  
+A: They are two different tools for avoiding parts of the code while stepping. skipFiles helps you skip stack frames that sit between frames that you are interested in (e.g. [https://sorbet.org/](sorbet)). finishFiles, on the other hand, help you avoid stack frames (and everything deeper than them) entirely.
+
+Q: Can I see an example?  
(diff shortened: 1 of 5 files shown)


jvilk-stripe:
@wingrunr21 I have previously reviewed this code internally and confirmed that it works. I believe this functionality has utility outside of Stripe:

* `skipFiles`: The primary utility within Stripe is to skip [Sorbet](http://sorbet.org)'s runtime type checking and other DSLs that interpose on method calls. This sort of pattern is common in other Ruby DSLs.
* `finishFiles`: The primary utility within Stripe is to avoid stepping thru autoloaded code. There are other autoloaders out there that would benefit from this feature (like [Zeitwerk](https://guides.rubyonrails.org/autoloading_and_reloading_constants.html)).

Within Stripe, these features make the stepping features of the debugger usable since otherwise engineers need to step through a lot of extraneous code.

jvilk-stripe:
@wingrunr21 friendly poke on this. I know you're busy. :) Let me know if there's anything we can do to make you feel better about this contribution!

PR #785, opened 2021-12-24
feat(erb): add HTML onEnterRules

hayesr:
This brings the same on-enter behavior from the HTML language package to ERB. Specifically, when the cursor is between a start and end tag, the editor inserts a new line, puts the cursor at the new indent, and moves the closing tag to the line below at the proper indent.

Currently ERB has this behavior:

After this change it works like this:

This is my first contribution here and I had problems running `yarn watch`. (There was an error with copying tree-sitter.wasm) However run and debug worked fine and I was able to test manually. I'm happy to add tests if someone can provide some direction.

- [ ] The build passes
- [x] TSLint is mostly happy
- [ ] Prettier has been run

### DIFF
--- a/packages/vscode-ruby/language-configuration-erb.json
+++ b/packages/vscode-ruby/language-configuration-erb.json
@@ -27,5 +27,29 @@
 		["`", "`"],
 		["<", ">"],
 		["%", "%"]
+	],
+	"onEnterRules": [
+		{
+			"beforeText": {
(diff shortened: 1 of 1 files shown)


dengdengyee:
I need this so much

dengdengyee (approved):

hayesr:
@dengdengyee @wingrunr21 Just want to ping you to avoid this being closed by the bot.

Issue #578, opened 2019-12-29
rubocop not using .rubocop.yml file


Zhorian:
### Your environment

- `vscode-ruby` version: 0.26.0
- Ruby version: 2.6.3p62
- Ruby version manager (if any): 1.29.9
- VS Code version: 1.41.1
- Operating System: Linux Mint 19.3
- Using language server? (eg `useLanguageServer` is true?) true

### Expected behavior
I'm expecting test/test_helper.rb to be ignored by rubocop based on my .rubocop.yml file. Confirmed config file is being used when running rubocop from the command line.

### Actual behavior
I am still receiving linting errors on the for test/test_helper.rb

*This and the next section should include screenshots, code samples, console output, etc. The more information we have to reproduce the better!*

settings.json
`{
    "editor.formatOnSave": true,
    "editor.formatOnSaveTimeout": 5000,
    "powermode.enabled": true,
    "powermode.enableShake": false,
    "ruby.useLanguageServer": true,
    "ruby.lint": {
        "rubocop": true
    },
    "ruby.format": "rubocop",
    "sync.gist": "d1b8c7400637948e668f4c6d174c5edb",
    "workbench.colorTheme": "Blueberry Banana",
    "workbench.iconTheme": "vscode-icons",
}`

.rubocop.yml
`AllCops:
  TargetRubyVersion: 2.6.3
  Exclude:
    - "**/*.erb"
    - db/schema.rb
    - "bin/**/*"
    - "node_modules/**/*"
    - test/test_helper.rb
    - test/channels/application_cable/connection_test.rb`

![image](https://user-images.githubusercontent.com/32066442/71561703-d5829f80-2a71-11ea-8a0f-dd625e5a6905.png)

Zhorian:
Interestingly this seems to occur when using [Solargraph](https://github.com/castwide/vscode-solargraph) too!

andrewmcodes:
Try turning the bundler option on. I’m not sure if the gem will automatically find the config file in your directory instead of looking globally. 

https://github.com/rubyide/vscode-ruby/blob/master/docs/linting.md#configuration-options

Zhorian:
I've updated my ruby lint settings to below and the issue persists.

`    "ruby.lint": {
        "rubocop": {
            "command": "rubocop",
            "useBundler": true,
        }
    },`

artm:
`"useBundler": true` is ignored when `"command"` is given.

artm:
# Context

[...]

PR #809, opened 2022-03-25
Use linter.command for formating too

ynsgnr:
Use linter.command for formating too so we can at least use our custom commands in an sh file, because currently its impossible to give arguments or config files to rubocop and its extremely annoying

Solves https://github.com/rubyide/vscode-ruby/issues/578

- [] The build passes
- [] TSLint is mostly happy
- [] Prettier has been run


wingrunr21:
Sorry, not merging this. It is not obvious that the linter configuration would affect formatting based on the configs. The format config needs to have custom command support added instead.

PR #810, opened 2022-03-28
Fix Build Process

manuth:
Currently, `yarn run build` won't run through if `packages/language-server-ruby/dist` directory doesn't exist.
Changes made in this PR will add a piece of code to fix this issue.

- [x] The build passes
- [x] TSLint is mostly happy
- [x] Prettier has been run

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -4,5 +4,6 @@
 	],
 	"npmClient": "yarn",
 	"useWorkspaces": true,
-	"version": "0.27.0"
+	"version": "0.27.0",
+	"concurrency": 1
 }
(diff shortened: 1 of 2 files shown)


manuth:
> The change to the esbuild config should not be necessary [per the docs](https://esbuild.github.io/api/#outdir)
> 
> Is this due to the lerna concurrency and the symlink steps into the plugin folder?

The esbuild-settings use `outfile` rather than `outdir`. Looks like the output directory isn't created automatically for the `outfile` setting.
If this is allowed by `esbuild`, specifying both `outfile` and `outdir` might be another solution.

hayesr:
I just tried this and was able to get it to build and run. I did run into some test problems, but maybe those should be for a different PR.

PR #842, opened 2022-12-12
Refresh some dependencies and refactor Github CI

chasestubblefield:
Since the upgrade to Node 16, the CI workflow has been failing, and needed updating. Some upgrades were also needed since they required an old version of `node-gyp`, which caused `yarn install` to fail on the latest Windows runner.

* Removed `semantic-release` and related packages. I can't find any evidence they are used anymore. Related to the `node-gyp` issue.
* Moved `vsce` out of root and into the two workspaces it is used in. Also upgraded to `@vscode/vsce`.
* Upgraded `lerna`. Our version was quite old, and relied on an old `node-gyp`.
* Refactored CI workflow. Upgraded actions to latest versions, and uses built-in yarn caching. Also the ruby version matrix for testing the language server is no longer needed, as it uses the prebuilt, checked-in `tree-sitter-ruby.wasm`.

I believe the current failures in CI now are true failures. Some diagnosing on Windows is required.

You can see the latest run from my fork here: https://github.com/chasestubblefield/vscode-ruby/actions/runs/3671826946

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,34 +6,20 @@ jobs:
     name: 'Test Language Server'
     runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
-        ruby: [2.4, 2.5, 2.6, 2.7]
     steps:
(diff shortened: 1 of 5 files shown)


chasestubblefield:
@Buzzkill-McSquare

